PR #4, opened 2015-07-06
Test runner: Add a "tests" subdir with an example unittest test case

mseaborn:
Change run_tests.py to look for tests in the "tests" subdir, and to
find unittests as well as doctests.

### DIFF
--- a/tests/utilities_test.py
+++ b/tests/utilities_test.py
@@ -1,0 +1,19 @@
+# Copyright 2015 The Switch Authors. All rights reserved.
+# Licensed under the Apache License, Version 2, which is in the LICENSE file.
+
+import unittest
+
+import switch_mod.utilities as utilities
+
+
+class UtilitiesTest(unittest.TestCase):
+
+    def test_approx_equal(self):
+        assert not utilities.approx_equal(1, 2)
+        assert not utilities.approx_equal(1, 1.02)
+        assert utilities.approx_equal(1, 1.01)
+        assert utilities.approx_equal(1, 1)
+
+
+if __name__ == '__main__':
+    unittest.main()
--- a/run_tests.py
+++ b/run_tests.py
@@ -8,18 +8,37 @@ import sys
 import unittest
 
 
-# The unittest module does not have built-in support for finding
-# doctests.  In order to run the doctests, we need a custom TestLoader
-# that overrides loadTestsFromModule().
 class TestLoader(unittest.TestLoader):
 
+    # unittest.main does not allow multiple "--start-directory"
+    # options, but we can make it scan multiple separate directories
+    # by overriding discover().  This allows us to have a "tests"
+    # directory that's separate from "switch_mod".
+    #
+    # We don't want to scan for *.py files in the parent directory in
+    # case any of those are throwaway scripts that have unexpected
+    # effects when imported.
+    def discover(self, start_dir, pattern, top_level_dir):
+        test_suite = unittest.TestSuite()
+        for subdir in ('switch_mod', 'tests'):
+            test_suite.addTests(
+                super(TestLoader, self).discover(
+                    os.path.join(top_level_dir, subdir),
+                    pattern, top_level_dir))
+        return test_suite
+
+    # The unittest module does not have built-in support for finding
+    # doctests.  In order to run the doctests, we need a custom
+    # TestLoader that overrides loadTestsFromModule().
     def loadTestsFromModule(self, module):
+        test_suite = super(TestLoader, self).loadTestsFromModule(module)
+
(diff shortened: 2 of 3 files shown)


tests/utilities_test.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class UtilitiesTest(unittest.TestCase):
+
+    def test_approx_equal(self):
+        assert not utilities.approx_equal(1, 2)
+        assert not utilities.approx_equal(1, 1.02)
+        assert utilities.approx_equal(1, 1.01)
+        assert utilities.approx_equal(1, 1)
+
```
josiahjohnston: I tried adding another few tests. The tests pass, but the error message from the exception is printed to the screen. I'd rather hide the expected error messages when testing code that is intended to detect and report errors by raising exceptions. 
```python
    def test_min_data_check(self):
        from pyomo.environ import AbstractModel, Set, Param
        mod = AbstractModel()
        utilities._add_min_data_check(mod)
        mod.set_A = Set(initialize=[1, 2])
        mod.paramA_full = Param(mod.set_A, initialize={1: 'a', 2: 'b'})
        mod.paramA_empty = Param(mod.set_A)
        mod.min_data_check('set_A', 'paramA_full')
        self.assertIsInstance(mod.create(), AbstractModel)
        mod.min_data_check('set_A', 'paramA_empty')
        # This prints the error message to the screen
        self.assertRaises(ValueError, mod.create)
        # So does this
        # with self.assertRaises(ValueError) as err:
        #     mod.create()
        # I don't know how to keep the error message from printing.
```

josiahjohnston:
Looks good :)
I still have a lot to learn about unittest, but this example was easy enough to extend slightly, even though I couldn't get it to behave completely.

tests/utilities_test.py
mseaborn: On 6 July 2015 at 17:43, Josiah Johnston <notifications@github.com> wrote:

> I tried adding another few tests. The tests pass, but the error message
> from the exception is printed to the screen. I'd rather hide the expected
> error messages when testing code that is intended to detect and report
> errors by raising exceptions.
>
>     def test_min_data_check(self):
>         from pyomo.environ import AbstractModel, Set, Param
>         mod = AbstractModel()
>         utilities._add_min_data_check(mod)
>         mod.set_A = Set(initialize=[1, 2])
>         mod.paramA_full = Param(mod.set_A, initialize={1: 'a', 2: 'b'})
>         mod.paramA_empty = Param(mod.set_A)
>         mod.min_data_check('set_A', 'paramA_full')
>         self.assertIsInstance(mod.create(), AbstractModel)
>         mod.min_data_check('set_A', 'paramA_empty')
>         # This prints the error message to the screen
>         self.assertRaises(ValueError, mod.create)
>         # So does this
>         # with self.assertRaises(ValueError) as err:
>         #     mod.create()
>         # I don't know how to keep the error message from printing.
>
> OK, I tried running that test, and I see what you mean.

I got the following output:

"""
$ ./run_tests.py
................................ERROR: Constructing component
'min_data_check_2' from data=None failed:
ValueError: Values are not provided for every element of the mandatory
parameter 'paramA_empty'
.
----------------------------------------------------------------------
Ran 33 tests in 1.367s

OK
"""


Searching for "Constructing component" in the Pyomo code shows what's
causing this:

$ rgrep 'Constructing component' deps/install/
deps/install/lib/python2.7/site-packages/Pyomo-4.0.9682-py2.7.egg/pyomo/core/base/PyomoModel.py:
               "Constructing component '%s' from data=%s failed:\n%s: %s",
deps/install/lib/python2.7/site-packages/Pyomo-4.0.9682-py2.7.egg/pyomo/core/base/block.py:
                   "Constructing component '%s' from data=%s failed:\n%s:
%s",

PyomoModel.py does this:

        try:
            declaration.construct(data)
        except:
            err = sys.exc_info()[1]
            logger.error(
                "Constructing component '%s' from data=%s failed:\n%s: %s",
                str(declaration.cname(True)), str(data).strip(),
                type(err).__name__, err )
            raise


Whether that is good practice is debatable.  It's useful because it adds
context to the exception.  Maybe it would be better to do that by
annotating or wrapping the exception object, though Python doesn't provide
a really good way to do that.

In any case, it doesn't interact well with unittest, because unittest does
not redirect stdout/stderr by default.  This is a longstanding omission,
and it has the unfortunate effect that if tests write to stdout or stderr,
we get messy output from unittest.

I can fix that by changing run_tests.py to redirect stdout/stderr during
each test.  Unfortunately unittest's test runner doesn't do enough out of
the box. :-(

Cheers,
Mark

PR #9, opened 2015-08-08
Change solve.py examples to use the GLPK solver instead of CPLEX

mseaborn:
This makes the examples runnable using only open source software --
since GLPK is open source (whereas CPLEX is proprietary).

Also move the choice of solver to one place in switch_mod, so that it
can be changed more easily.

### DIFF
--- a/examples/3zone_toy/solve.py
+++ b/examples/3zone_toy/solve.py
@@ -16,15 +16,15 @@ directory is in your python search path. See the README for more info.
 """
 
 from pyomo.environ import *
-from pyomo.opt import SolverFactory
 from switch_mod.utilities import define_AbstractModel
+import switch_mod.utilities
 
 switch_model = define_AbstractModel(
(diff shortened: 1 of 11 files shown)


josiahjohnston:
The code looks good to me. If the tests pass for you, I say commit it.

mseaborn:
On 7 August 2015 at 19:02, Josiah Johnston <notifications@github.com> wrote:

> The code looks good to me. If the tests pass for you, I say commit it.
>
Tests pass in the sense that all the solve.py scripts run without giving
any errors.  And run_tests.py still passes too, of course.

There are some differences from these two checked-in output files:
examples/3zone_toy/outputs/dispatch.txt
examples/3zone_toy/outputs/load_balance.txt

-- some floating-point numbers differ by small amounts.

Presumably that is a difference between CPLEX and GLPK, but since I haven't
got a copy of CPLEX I can't check.  I'll resolve this when I turn the
solve.py scripts into tests. :-)

Cheers,
Mark

PR #10, opened 2015-09-24
Add a command line front end for running Switch: switch_mod.solve

mseaborn:
This tool takes the list of modules to use from an "inputs/modules"
file.  I have added "modules" files to each of the examples (except
for "custom_extension", for now).

Note that this duplicates the module lists in the examples' solve.py
files.  In a later change I'll remove those solve.py files, but move
their docstring comments to separate readme files.

I tested this by running all the examples (except "custom_extension").


josiahjohnston:
Looks good to me. 
It would be nice if solve.py would print a useful error message if inputs/modules file is missing instead of crashing with a traceback.

mseaborn:
Thanks for the review!  I'm addressing the error handling with a separate change: https://github.com/switch-model/switch/pull/11

PR #15, opened 2016-02-13
Run all the examples as tests

mseaborn:


josiahjohnston:
Functionality looks good.

I just noticed a pre-existing bug: the test framework has a dependency on testfixtures, but this isn't mentioned in the INSTALL file. It crashes if that python module isn't installed. Could you add a note to INSTALL as you merge things in?

mseaborn:
On 12 February 2016 at 18:42, Josiah Johnston <notifications@github.com>
wrote:

> Functionality looks good.
>
OK, I'll take that as an "LGTM" and commit this. :-)

> I just noticed a pre-existing bug: the test framework has a dependency on
> testfixtures, but this isn't mentioned in the INSTALL file. It crashes if
> that python module isn't installed. Could you add a note to INSTALL as you
> merge things in?
>
I noticed that too, since I had to install testfixtures.  I will do a
separate change to update the instructions for installing dependencies.

Cheers,
Mark

PR #16, opened 2016-03-05
Use pinned versions for Pyomo and other Python library dependencies

mseaborn:
Josiah, can you check that that "pip install" command works OK on your Mac?

Cheers,
Mark


mfripp:
Glad to see this happening -- I was gearing up to do it myself! 

By the way, pip runs very badly with the system-installed version of Python on El Capitan (even with --user). I figured out a bunch of workarounds and posted them at http://apple.stackexchange.com/a/223163 but I would say it's definitely best to use a separate Python installation now.

mseaborn:
Josiah, can you let me know whether I can commit this?  If you haven't got time to test this, can I commit it anyway, and we can change it if we later find problems with it?

josiahjohnston:
Yes, you can commit this and we can fix it later. In its current form, this is as good or better than what we had before the commit.
I’m sorry for the delay. I’ll test it on a fresh mac computer when I get a chance.

-Josiah

PR #17, opened 2016-03-16
Advanced argument parsing and multi-scenario scripting

mfripp:
This commit includes a new version of solve.py and a new solve_scenarios.py. 

solve.py defines a number of solver-related command-line arguments, reads them from options.txt and/or the command line, creates a model instance, customizes a solver based on the command-line arguments, and solves the model. It can also solve models repeatedly in order to iterate to convergence or conduct sensitivity studies. Argument processing for solve.py is integrated with argument processing for add-on modules (e.g., switch_hawaii.demand_response.py in the hawaii_dr branch), so this provides a complete interface for customizing model runs.

solve_scenarios.py reads scenario definitions from scenarios.txt (or similar) and then sends the scenario definitions to solve.py. The scenario definitions are just lists of command-line arguments -- the same arguments can be used in options.txt, scenarios.txt or on the command line. Some of these affect the solver, and some of them are consumed by add-on modules. solve_scenarios.py also includes code to allow single or parallel instances to work their way through the scenario list without duplicating effort (and even to respond to changes in the scenario list while scenario_solve is running).

Some notes on solve.py:

- solve.py uses the new argument-parsing capabilities in utilities.py pretty extensively. 
- It reads any command-line arguments in options.txt and then appends any arguments specified on the actual command line.
- It reads a list of modules from modules.txt (or from a different file specified by `--module_list <listfile>`). It also appends any modules specified with `--include_module <mod>` or `--include_modules <mod1> <mod2> ...`, and then it excludes modules specified with `--exclude_module <mod>` or `--exclude_modules <mod1> <mod2> ...`. 
- Note that arguments specified on the command line will replace any specified earlier (e.g., in options.txt). So solve.py always starts with the list in modules.txt (or the file specified by the last --module_list command), then applies the _last_ --include_module(s) command, then the _last_ --exclude_module(s) command. This seems to give enough flexibility and predictability.
- It passes the module list to switch_mod.utilities.create_model() (similar to define_AbstractModel) to create a model.
- Note that code added to switch_mod.utilities.create_model() in a previous commit creates an argument parser and calls define_arguments(argparser) for all modules in the list. Any module can include this function to define command-line arguments that it wants to consume. Then switch_mod.utilities.create_model() parses all arguments and places the results in model.options (e.g., model.options.inputs_dir). This provides a flexible system for passing options into modules as needed.
- solve.py registers itself in the module list and uses the standard approach to define a number of solver-related arguments (see define_arguments())
- solve.py builds a customized solver based on options specified in options.txt and the command line. It adds this to the model as model.solver. This supports the possibility for creating a model.solve() method later or for defining two different models at the same time with different solver options. 
- The solve() function in solve.py calls solver.solve() with a customized list of arguments, including a list of suffixes created by other modules.
- solve.py also defines automatic iteration behavior. This is useful for models that must iterate to convergence (e.g., with nonlinear demand response functions and/or AC power flow). It could also potentially be used for sensitivity studies (e.g., carbon_costs=[10, 100, 1000]; model.carbon_cost=carbon_costs[model.iteration_number]). If modules are listed in iterate.txt, then solve.py will repeatedly call their pre_iterate() function, then solve(), then call their post_iterate() function to test for convergence/completion. If modules are listed on different lines in iterate.txt, then they define multi-level iteration loops. If they are listed on the same line, then they will all be tested for convergence at the same level.

Some notes on scenario_solve.py:

- scenario_solve.py reads a list of command line options for different scenarios from scenarios.txt or another file specified by `--scenario_list <scenario_file>`. It passes each of these lists of arguments in turn to the solve module. 
- one of the arguments passed to the solve module is scenario_name, which can be used by modules to give their output files unique names or store them in unique directories. 
- scenario_solve can be called with a `--scenario` or `--scenarios` option to choose one or more scenarios from the list; otherwise it will run all scenarios in the list.
- when running all scenarios in the list, scenario_solve will check whether each one has already been solved; if so, it will move onto the next scenario. After each solve, it will check the whole list again from the top. This allows users to add new scenarios to the list, and scenario_solve will solve them without being restarted.
- the code to check whether a scenario has run is designed to be multi-process safe on networked file systems. This means many copies of scenario_solve can be run simultaneously on the same computer or on different computers on a networked file system, and they will allocate the scenarios among themselves until all are completed. scenario_solve.py uses "lock directories" for each scenario, stored under scenario_queue/ (or another location specified by `--scenario_queue <dir>`). Scenarios can be re-queued by removing the empty directory with the same name. Or the whole queue can be restarted by executing `rm -rf scenario_queue`.
- when it starts, scenario_queue.py tries to create a unique, persistent job id. On HPC environments, this will be derived from the SLURM or MPI job ID. On ad hoc clusters, you may want to specify `--job_id <x>` for each separate worker. When it launches, if scenario_queue finds any scenarios that were previously being solved by the same job but got interrupted, it will requeue them. If you have a large sequence of scenarios to run, this allows you to interrupt and restart jobs without losing any scenarios. Scenarios that were previously being processed by this job are listed in scenario_queue/<job_id>_running.txt. I haven't found a way to re-queue the running jobs automatically when scenario_solve.py is interrupted. They will only be re-queued if you restart scenario_solve with the same job id or manually remove the lock directory for that scenario.

Take a look at the hawaii_dr branch (especially options.txt, iterate.txt, scenarios.txt and/or scenarios_tiny.txt, and switch_hawaii.demand_response) to see how these pieces work together. You can just run `python -m switch_mod.solve` to solve the default scenario defined in options.txt and iterate.txt, or run `python -m switch_mod.solve_scenarios --scenario_file scenarios_tiny.txt` to solve a small list of scenarios. You can also add command line arguments to either of these, e.g., `--biofuel_limit 0.05` or `--suffix iis`.

These scripts also provide help if called with `--help`. If a module list is defined, `python -m switch_mod.solve --help` will also give you help messages from all the modules that are loaded.

### DIFF
--- a/switch_mod/solve.py
+++ b/switch_mod/solve.py
@@ -1,62 +1,398 @@
-# Copyright 2015 The Switch Authors. All rights reserved.
-# Licensed under the Apache License, Version 2, which is in the LICENSE file.
+#!/usr/bin/env python
 
-"""
-Command line front-end for running the Switch model solver.
+import sys, os, time, traceback, shlex, re
 
-Usage:  python -m switch_mod.solve [ARGS]
-"""
+from pyomo.environ import *
+from pyomo.opt import SolverFactory, SolverStatus, TerminationCondition
 
-import argparse
-import os
-import sys
+def add_relative_path(*parts):
+    """ Adds a new path to sys.path.
+    The path should be specified relative to the current module,
+    and specified as a list of directories to traverse to reach the
+    final destination."""
+    sys.path.append(os.path.abspath(os.path.join(os.path.dirname(__file__), *parts)))
 
-import pyomo.opt
+add_relative_path('switch')   # standard switch model
+from switch_mod.utilities import create_model, _ArgumentParser
 
-import switch_mod.utilities
+add_relative_path('.') # components for this particular study
 
+def main(args=sys.argv[1:]):
+        
+    # build a module list based on configuration options passed in,
+    # and add the current module (to register define_arguments callback)
+    modules = get_module_list(args)
+    
+    # NOTE: we have (at least) two options for parsing model configuration arguments 
+    # (these may come from option files or the command line):
+    # 1. try to do all the parsing in solve.py and leave utilities.py ignorant of the argument framework
+    # 2. make command-line arguments a core part of the modeling framework in utilities.py
+    # I opted for option 1 because
+    # - each module may need to define its own arguments (e.g., EV scheduling rule or result storage request)
+    # - this means we need to call each module to define arguments and then parse the arguments
+    #   after creating the AbstractModel and before calling define_components. That can only be
+    #   done in utilities.py
+    # - parsing command line arguments in utilities.py allows us to define standard options
+    #   for the solver (and create a standard solve function in utilities.py) and for 
+    #   input/output directories (eliminating the need to pass variables around for these).
+    # - defining an options object (based on the arguments) in the utilities.py allows us to guarantee
+    #   that these elements will be available no matter what solution technique is used (standard
+    #   solve.py or runph or parallel solver)
 
-def main(argv):
-    parser = argparse.ArgumentParser(
-        prog='python -m switch_mod.solve',
-        description='Runs the Switch power grid model solver.')
-    parser.add_argument(
-        '--inputs-dir', type=str, default='inputs',
+    # Define the model
+    model = create_model(modules, args=args)
+
+    # Add any suffixes specified on the command line (usually only iis)
+    add_extra_suffixes(model)
+    
+    # get a list of modules to iterate through
+    iterate_modules = get_iteration_list(model)
+    
+    print "\n\n======================================================================="
+    print "arguments:"
+    print " ".join(k+"="+repr(v) for k, v in model.options.__dict__.items() if v)
+    print "modules:", modules
+    if iterate_modules:
+        print "iterate_modules", iterate_modules
+    print "======================================================================="
+
+    # create an instance
+    instance = model.load_inputs()
+
+    # make sure the outputs_dir exists (used by some modules during iterate)
+    if not os.path.exists(instance.options.outputs_dir):
+        os.makedirs(instance.options.outputs_dir)
+
+    # solve the model
+    if iterate_modules:
+        print "iterating model..."
+        iterate(instance, iterate_modules)
+    else:
+        print "solving model..."
+        solve(instance)
+        print "finished solving"
+    
+    # report/save results
+    instance.post_solve()
+
+def iterate(m, iterate_modules, depth=0):
+    """Iterate through all modules listed in the iterate_list (usually iterate.txt),
+    if any. If there is no iterate_list, then this will just solve the model once.
+    
+    If it exists, the iterate_list contains one row per level of iteration,
+    and each row contains a list of modules to test for iteration at that level
+    (these can be separated with commas, spaces or tabs).
+    The model will run through the levels like nested loops, running the lowest level 
+    till it converges, then iterating the next higher level by one step, then running the
+    lowest level to convergence again, repeating until all levels are converged.
+    During each iteration, the pre_iterate() and post_iterate() functions of each specified
+    module (if they exist) will be called before and after solving. When a module is 
+    converged, its post_iterate() function should return True. 
+    All modules specified in the iterate_list should also be loaded via the module_list
+    or include_module(s) arguments.
+    """
+    
+    # create or truncate the iteration tree
+    if depth == 0:
+        m.iteration_node = []
+
+    if depth == len(iterate_modules):
+        # asked to converge at the deepest level
+        # just preprocess to reflect all changes and then solve
+        m.preprocess()
+        solve(m)
+    else:
+        # iterate until converged at the current level
+
+        # note: the modules in iterate_modules were also specified in the model's 
+        # module list, and have already been loaded, so they are accessible via sys.modules
+        current_modules = [sys.modules[module_name] for module_name in iterate_modules[depth]]
+        # truncate the iteration tree at the current level
+        m.iteration_node = m.iteration_node[:depth] + [0]
+
+        j = 0
+        converged = False
+        while not converged:
+            # take one step at the current level
+            if m.options.max_iter is not None and j >= m.options.max_iter:
+                break
+
+            # record the current iteration number and node for use by modules 
+            # (e.g., to name files or reset/index params)
+            m.iteration_number = j
+            m.iteration_node[-1] = j
+
+            converged = True
+            # pre-iterate modules at this level
+            for module in current_modules:
+                if hasattr(module, 'pre_iterate'): 
+                    converged = module.pre_iterate(m) and converged
+
+            # converge the deeper-level modules, if any (inner loop)
+            iterate(m, iterate_modules, depth=depth+1)
+            
+            # post-iterate modules at this level
+            for module in current_modules:
+                if hasattr(module, 'post_iterate'):
+                    converged = module.post_iterate(m) and converged
+
+            j += 1
+        if converged:
+            print "Iteration of {ms} converged after {j} rounds.".format(ms=iterate_modules[depth], j=j)
+        else:
+            print "Iteration of {ms} was stopped after {j} iterations without convergence.".format(ms=iterate_modules[depth], j=j)
+    return
+
+def define_arguments(argparser):
+    # callback function to define model configuration arguments while the model is built
+
+    # add standard module arguments (not used later, but this adds them to the help)
+    add_module_args(argparser)
+
+    # iteration options
+    argparser.add_argument(
+        "--iterate_list", default=None,
+        help="Text file with a list of modules to iterate until converged (default is iterate.txt); "
+             "each row is one level of iteration, and there can be multiple modules on each row"
+    )
+    argparser.add_argument(
+        "--max_iter", type=int, default=None,
+        help="Maximum number of iterations to complete at each level for iterated models"
+    )
+
+    # scenario information
+    argparser.add_argument(
+        "--scenario_name", default="", help="Name of research scenario represented by this model"
+    )
+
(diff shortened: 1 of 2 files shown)


mfripp:
By the way, the path-setting code at the top of solve.py and solve_scenarios.py is obsolete and should be removed.

switch_mod/solve.py
```diff
@@ -1,62 +1,398 @@
...
-import switch_mod.utilities
+add_relative_path('.') # components for this particular study
 
+def main(args=sys.argv[1:]):
+        
+    # build a module list based on configuration options passed in,
+    # and add the current module (to register define_arguments callback)
+    modules = get_module_list(args)
+    
+    # NOTE: we have (at least) two options for parsing model configuration arguments
```
mfripp: You can ignore these notes (we should delete most of them soon). The approach I ended up adopting was to put  minimal support for arguments in switch_mod.utilities (create an argument parser, pass it to the modules, parse the arguments and put the results in model.options), and then have each module define and consume the arguments it needs. So switch_mod.utilities doesn't actually define or operate on any arguments, but it provides a skeleton framework which makes it easy to use them everywhere.

switch_mod/solve.py
```diff
@@ -1,62 +1,398 @@
...
-import sys
+def add_relative_path(*parts):
+    """ Adds a new path to sys.path.
+    The path should be specified relative to the current module,
+    and specified as a list of directories to traverse to reach the
+    final destination."""
+    sys.path.append(os.path.abspath(os.path.join(os.path.dirname(__file__), *parts)))
 
-import pyomo.opt
+add_relative_path('switch')   # standard switch model
```
mfripp: All the path-setting code here is obsolete and should be removed.

switch_mod/solve_scenarios.py
```diff
@@ -0,0 +1,281 @@
...
+long solver jobs are running. Multiple solvers can also use scenarios_to_run()
+in separate processes to select the next job to do.
+"""
+
+from __future__ import print_function
+import sys, os, time
+import argparse, shlex, socket, io, glob
+from collections import OrderedDict
+
+def add_relative_path(*parts):
```
mfripp: The path-setting code here is obsolete and should be removed.

josiahjohnston:
This has a lot of nice new features :)
The downside is that it changes the module list extensively enough to break all existing examples, which comprise 10 of the 43 tests (execute from root directory with ./run_tests.py). The problematic changes are:
* Changing the default path from input_dir/modules to modules.txt
* Not adding switch_mod to the beginning of the module list

I started to write some new code into get_module_list() to be backwards compatible with existing module lists, but it got ugly quickly because the path to the inputs directory is not available inside of the get_module_list() function based on how you split argument additions across define_arguments() and add_module_args(). I could merge all of those arguments into define_arguments(), but I don't know if there was a reason for splitting them.

The other solution is to redo the modules file of each example to match the new format and expectations. This is not ideal from my perspective because I think of the list of modules are part of the inputs, but is workable for merging things back in. I verified that multiple examples produce identical results after I edit and rename the modules file. 

Do you have a preference of how to proceed? I'm guessing the latter option may be preferable because it would entail fewer changes for your run directories..

mfripp:
Glad you like it. I think there is a fundamental design choice to be made here -- whether the model configuration files should be inside the inputs directory or alongside it. I would vote for the latter.

I often run a "study", which consists of a set of scenarios which use mostly the same modules and options, but draw from several different input directories. I may also have directories with much smaller datasets to use for testing. All of these are generated via automated scripts. In this context, it makes sense to place the module list, options file and scenarios file as siblings of the inputs directories, rather than putting them inside the inputs directory. This looks something like the following:

study_1/	# all files needed for one particular set of scenarios
    modules.txt
    options.txt
    scenarios.txt
    inputs_2045/      # inputs for some scenarios
        data_file_1.tab
        data_file_2.tab
        data_file_3.tab
    inputs_2007/      # inputs for other scenarios
        data_file_1.tab
        data_file_2.tab
        data_file_3.tab
    inputs_tiny/      # inputs for testing
        data_file_1.tab
        data_file_2.tab
        data_file_3.tab

I don't really see any way to achieve the same functionality with model configuration files stored inside the inputs directories. In particular, scenarios.txt may list a collection of similar scenarios for comparison, some using the 2045 data and some using the 2007 data (or some using high prices and some using low prices, etc.). Multiple scenarios may use the same inputs directory, so if I wanted to store the scenario definitions inside the inputs directories, I would have to break scenarios.txt into separate slices for each directory. But then I'd need a meta-scenario list at a higher level to identify which scenarios I want to run right now (and where to find them). I could put identical copies of modules.txt into each inputs directory, but that would be seriously error-prone (I might change modules.txt and options.txt in the inputs_tiny directory while testing, but then forget to make the same changes in the other directories before running the final model).

Even when there is only one directory of inputs, I think this is a better design, because it keeps the model configuration files from getting mixed in with the data files. This reflects the idea of using a generic model with alternative data inputs. It is also easier to edit and maintain, because users can easily find the configuration files without having to sort through the data files. I often need to edit the configuration files, but rarely edit the data files by hand (and in fact I may occasionally regenerate the data files via an automated script).

If you want to achieve backward compatibility with the current approach, that could be done in solve.py. That module pre-parses the arguments, just enough to find the module list to pass to utilities.create_model(). It would be easy to change solve.py to pre-parse the "--inputs-dir" argument, and then check that directory for a "modules" file if it doesn't find a "modules.txt" in the outer directory. But I don't think it would be useful to have this capability, and offering two methods may confuse users about the "right" way to setup a model.

About the automatic pre-pending of the switch_mod package name: I think I decided that since I needed to be able to choose between modules in switch_mod and modules in switch_hawaii, it made sense to specify explicitly which module I was trying to load (just as I would with an import statement). Automatically adding a package name doesn't save much effort and could confuse people about what is going on. I think this also made the import code slightly simpler, and may have had some implications for modules placed loose in the local directory with the same names as standard switch modules. I would however be in favor of renaming the modules to something like switch.fuel_cost, switch.commit, switch.hawaii.demand_response, etc. Then maybe we could go back to automatically dropping the prefix...

Matthias

mfripp:
I just updated this branch with code to search for inputs_dir/modules if modules.txt is not found. We should still decide whether to use ./modules.txt as our standard instead, but this should get you through the tests.

I misunderstood your comment about adding switch_mod to the start of the modules list. You can disregard my earlier comment. I have now added code to insert switch_mod at the start of the module list if the user doesn't specify it elsewhere.

mfripp:
I have also added a new setup.py module which creates a "switch" command-line script. 

I expect to use "python setup.py develop" often to add the current development version of switch_mod (and the switch command) to the system paths. This appears to be the easiest way to use an under-development copy of switch to solve a model that is not stored within the same repository. e.g., I will have directory structures like this:

study/    <- repository shared with colleagues
    modules.txt
    config.txt
    inputs1/
    inputs2/
    ...
    switch/     <- switch repository (submodule) to be used with this study

In this setup, the appropriate copy of switch can be activated via "cd switch; python setup.py develop". Then "python -m switch_mod.solve" or "switch solve" will solve the model in the current directory.

A similar approach could be used with this directory structure:

switch/   <- latest development version of switch
study1/    <- repository shared with colleagues
    modules.txt
    config.txt
    inputs1/
    inputs2/
    ...
study2/    <- repository shared with colleagues
    modules.txt
    config.txt
    inputs1/
    inputs2/
    ...

It may not be a good idea to use setup.py if you are using the following directory structure (which I think may be what you are usually doing):

study1/
    switch/
        switch_mod/
        modules.txt
        config.txt
        inputs1/
        inputs2/
        ...
study2/
    switch/
        switch_mod/
        modules.txt
        config.txt
        inputs1/
        inputs2/
        ...

In this case, you can always solve correctly by running "python -m switch_mod.solve" inside the appropriate switch directory. If you run "python setup.py develop", you will get the handy "switch" command-line script, but this script will point to one particular copy of switch. So if you cd into a different study, you could end up running the wrong copy of switch. 

The first two structures can also lead to running the wrong copy of switch, but for those structures we have to use something like setup.py to get switch_mod into the path. And those structures are attractive because they allow easy sharing of the model configuration and data without having to make every study a separate branch of switch. The fundamental challenge is this: if module configuration files are siblings of switch_mod, then they cannot easily be shared via a separate repository. But if they are not siblings, then there is no natural way to have the model configuration files and switch_mod in the path at the same time. So we put switch_mod in the path via setup.py. 

(This approach is also compatible with other solutions: users who aren't modifying switch itself can just run "pip install switch"; users who want a more comprehensive environment-switching solution can use a virtualenv and run "setup.py develop" in there.)

mfripp:
Just added a couple more commits so run_tests.py succeeds with the new switch_mod.solve module. It gives deprecation warnings about the location of the modules files, but that can be fixed by moving the files or removing the warning code.

mfripp:
FYI, if you want a fairly simple model to test this out on, you can use the repository at https://github.com/switch-hawaii/ge_validation . This also gives an example of how a shared repository can be created for a single study. My grad student and I will eventually use that repository for validation work, but for now it just shows a fairly generic switch-hawaii model. 

You can change the --input-dir in options.txt or on the command line to switch between a tiny, forward-looking model (inputs_tiny), or a backward-looking production cost model (inputs_2007_15) or a larger forward-looking model (inputs_2045_15). For the 2007 one, you should also specify "--exclude-module hawaii.rps".

mfripp:
I'm ready to merge this pull request if there are no objections. I have promoted inputs/modules.txt to be a standard location for the module list, along with modules.txt in the main directory. To support this, I have also renamed all the inputs/modules test files to inputs/modules.txt. If anyone has their module list in inputs/modules, they will get a deprecation warning during solve. (I think it would be confusing to have inputs/modules be a standard location along with modules.txt.)

mfripp:
I'm ready to merge this pull request if there are no objections. This will treat both inputs/modules.txt and modules.txt as standard locations for the module list. To support this, this branch also renames inputs/modules to inputs/modules.txt in the examples and test_dat directories. It will give a deprecation warning during solve if you use inputs/modules, because I think it would be confusing to have both modules.txt and inputs/modules as standard locations for the module list.

PR #18, opened 2016-04-14
Fixing some things in the transmission modules

bmaluenda:
There was a conflict in the calculation of the LZ_TXNet expression between the arguments of the function and the indexes of the for loop.
Transmission capacity was calculated using the derating factor as the availability, when the availability is actually (1 - derating factor).

Also changed the name of a parameter in the trans_params.dat documentation to match the name of the Pyomo parameter.

### DIFF
--- a/switch_mod/trans_build.py
+++ b/switch_mod/trans_build.py
@@ -236,7 +236,7 @@ def define_components(mod):
     mod.TransCapacityAvailable = Expression(
         mod.TRANSMISSION_LINES, mod.PERIODS,
         rule=lambda m, tx, period: (
-            m.TransCapacity[tx, period] * m.trans_derating_factor[tx]))
+            m.TransCapacity[tx, period] * (1 - m.trans_derating_factor[tx])))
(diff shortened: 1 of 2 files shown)


bmaluenda:
Merging some bug fixes and very minor corrections in documentation.

josiahjohnston:
Thanks for spotting and squishing these bugs!

trans_derating_factor will be more consistent with Switch-AMPL and easier to use (I think) if we treat this fudge factor like efficiency so that a value of 1 has no effect and a value of 0 disables the line completely.

Please do the following: 
Update the documentation of trans_derating_factor to explain the intended use.
Update the default value of trans_derating_factor to be 1 instead of 0.
Revert the equation that uses trans_derating_factor to reflect its intended meaning.
Fix the syntax error on the correction to trans_cost_annual

Also, please update the outputs in 3zone_toy. Good call on that!

-Josiah

bmaluenda:
The corrections were made and commited to the master branch. Closing this pull request now.

Thanks for the comments!

PR #20, opened 2016-05-08
Test the objective function value (total cost) generated for the examples

mseaborn:
switch_mod/utilities.py: Add the ability to export the value of the
objective function to total_cost.txt.

Check in expected values for the examples' total costs in the
"examples" directory.

Change examples_test.py to test for those expected values, with an
option ("--update") for updating them easily whenever changes to the
model require this in future.


josiahjohnston:
Looks great and works for me! 👍

PR #23, opened 2016-06-04
Cleanup: De-duplicate copying of optional_params in load_aug()

mseaborn:
The same copying is done earlier in the function.


josiahjohnston:
Looks good to me.

Issue #24, opened 2016-06-14
Switch doesn't allow having zero existing builds


mseaborn:
Switch currently requires there to be at least one existing build.

Steps to reproduce: Take `examples/copperplate0` and modify it by deleting all rows from `proj_existing_builds.tab` and `proj_build_costs.tab`.  Run switch_mod.solve.

Expected result: No error.

Actual result: Switch gives the following error:

```
ERROR: Constructing component 'proj_existing_cap' from data={None: 'PROJECT'} failed:
	RuntimeError: Failed to set value for param=proj_existing_cap, index=None, value=PROJECT.
		source error message="Error setting parameter value: Index 'None' is not valid for array Param 'proj_existing_cap'"
Traceback (most recent call last):
...
  File ".../pyomo/core/base/param.py", line 802, in construct
    % (self.cname(True), str(key), str(val), str(msg)) )
RuntimeError: Failed to set value for param=proj_existing_cap, index=None, value=PROJECT.
	source error message="Error setting parameter value: Index 'None' is not valid for array Param 'proj_existing_cap'"
```

I debugged this and I discovered that Pyomo has a dubious special case for 1-line `.tab` files.  (See the `elif len(tmp) == 1` case in `pyomo/core/plugins/data/text.py`.)

I would expect Pyomo to treat a 1-line file as containing no rows of data (just a row of headings).  However, instead Pyomo seems to be converting the file to the declaration `param proj_existing_cap := PROJECT`.

A possible fix would be to change `load_aug()` in `utilities.py` so that it skips any `.tab` files that contain zero rows.  `load_aug()` already has a check for that, but it's only enabled when it's called with `optional=True`.

bmaluenda:
This was fixed by you in [this commit](https://github.com/switch-model/switch/commit/3b77568299505652c7e248011ea883543ab1b907) from August 17th, right Mark?

Could we close this Issue then? (I'm trying to clean up a bit) 
Sorry for not commenting when it was relevant, I missed this post.

mseaborn:
Yes, this is fixed -- closing.  Thanks for the reminder.

PR #25, opened 2016-06-14
Make the handling of empty .tab files a bit stricter

mseaborn:
Previously, if the first or second lines were empty, the whole of the
file was skipped (for files loaded with optional=True), even if more
lines were present.

Instead, only skip the file if it contains 0 or 1 lines.  Validate the
headings if they are present.

### DIFF
--- a/switch_mod/utilities.py
+++ b/switch_mod/utilities.py
@@ -620,10 +620,13 @@ def load_aug(switch_data, optional=False, auto_select=False,
     optional_params=list(optional_params)
     # Parse header and first row
     with open(path) as infile:
-        headers = infile.readline().strip().split('\t')
-        dat1 = infile.readline().strip().split('\t')
-    # Skip if the file is empty or has no data in the first row.
-    if optional and (headers == [''] or dat1 == ['']):
+        headers_line = infile.readline()
+        second_line = infile.readline()
+    file_is_empty = (headers_line == '')
+    file_has_no_data_rows = (second_line == '')
+    headers = headers_line.strip().split('\t')
+    # Skip if the file is empty.
+    if optional and file_is_empty:
         return
     # Try to get a list of parameters. If param was given as a
     # singleton or a tuple, make it into a list that can be edited.
@@ -698,6 +701,10 @@ def load_aug(switch_data, optional=False, auto_select=False,
         for (i, p_i) in del_items:
             del kwds['select'][i]
             del kwds['param'][p_i]
+    if optional and file_has_no_data_rows:
+        # Skip the file.  Note that we are only doing this after having
+        # validated the file's column headings.
+        return
     # All done with cleaning optional bits. Pass the updated arguments
     # into the DataPortal.load() function.
     switch_data.load(**kwds)


josiahjohnston:
Validating column headers before skipping files where the first data row is empty sounds fine. That feature might be useful for people who are gradually assembling files, and I don't see it having a strong downside.

If the first data line is blank, but subsequent data lines are not blank, that sounds like an error condition. I don't see anything in the code that would avoid skipping the file if it contained more than 1 line. Can you elaborate and/or provide a description of the use case you are working with?

The code itself looks fine from my first read, but I don't understand how it relates to, "only skip the file if it contains 0 or 1 lines."

mseaborn:
On 21 June 2016 at 09:47, Josiah Johnston <notifications@github.com> wrote:

> Validating column headers before skipping files where the first data row
> is empty sounds fine. That feature might be useful for people who are
> gradually assembling files, and I don't see it having a strong downside.
>
> If the first data line is blank, but subsequent data lines are not blank,
> that sounds like an error condition.
>
I don't see anything in the code that would avoid skipping the file if it
> contained more than 1 line.
>
The subtlety is that readline() returns "\n" for an empty line but "" at
the end of a file.

So, if you do:
    open(filename).readline().strip() == ""
that returns whether the first line is empty (or contains only whitespace).

If you do:
    open(filename).readline() == ""
that returns whether the file is completely empty.

This change is changing the code from a check like the former to a check
like the latter.

Can you elaborate and/or provide a description of the use case you are
> working with?
>
The background is this:  https://github.com/switch-model/switch/issues/24
-- "Switch doesn't allow having zero existing builds".

I guess I should have referenced that in the commit message, because you
might not have noticed that I created that issue.

This change does not fix that issue.  It's a cleanup which will make it
easier to fix that issue without making Switch too permissive about
accepting oddly-formatted input files.


> The code itself looks fine from my first read, but I don't understand how
> it relates to, "only skip the file if it contains 0 or 1 lines."
>
We skip the file if "optional and file_is_empty" is true (when the file has
0 lines) or when "optional and file_has_no_data_rows" is true (when the
file has 1 lines).

Cheers,
Mark

josiahjohnston:
Can you change the commit message?

It seems like the file would also be skipped if it contained numerous empty lines, or if it starts with blank lines before the headers or data starts (which seems like appropriate behavior). Other than the commit message, the more robust code seems good, and I'm fine with checking the headers before skipping a no-data condition.

mseaborn:
On 21 June 2016 at 11:13, Josiah Johnston <notifications@github.com> wrote:

> Can you change the commit message?
>
> It seems like the file would also be skipped if it contained numerous
> empty lines, or if it starts with blank lines before the headers or data
> starts (which seems like appropriate behavior).
>
That's not the case, actually.  With this change, the file won't be skipped
if it contains numerous empty lines or if the headings line isn't the first
line.

Or did you mean 'should' rather than 'would'?

Arguably the code should skip past any empty lines (and lines containing
only whitespace) when doing this check, to match more closely
how pyomo/core/plugins/data/text.py processes the .tab files, since that
also skips empty lines.  I am planning to change it to match pyomo's
text.py, but in a separate follow-on change because it makes the code more
complex.  Is that OK?

Cheers,
Mark

josiahjohnston:
I didn't realize that pyomo's .tab plugin had that behavior. Ignoring blank lies seems like a bug to me since I expect them to indicate an error in data file generation. Matching their behavior (feature or bug) makes sense in a way, even though this change does not fully match their behavior. Please update the commit message to explain why we are introducing this seemingly buggy behavior. 

Also note that empty lines before the error can still generate an error (I added blank lines at the beginning of project_info.tab in copperplate0 to test).

josiahjohnston:
This looks good to me if you update the commit message. Cheers :)

josiahjohnston:
Never mind about the commit message, this looks good as-is. I think I got confused about the "\n" vs "" issue earlier (even after you explained it), and then got confused further and thought you said this code was mimicking the screwy behavior of pyomo/core/plugins/data/text.py that ignores empty lines.

While I was poking at it tonight and making up test cases, the code behavior matches your commit message exactly, and all the standard tests still pass, so I'm fine with it. Whatever issues I thought I saw before, I can't recreate them.

PR #26, opened 2016-06-29
Allow models with no existing builds

mseaborn:
Make the input file "proj_existing_builds.tab" optional.

Issue: https://github.com/switch-model/switch/issues/24


josiahjohnston:
Looks good but I don't have time to test it now.
Benjamin, if you have time, could you test that this runs ok?
Thanks,
Josiah

bmaluenda:
I tried some tests and it works ok. I could replicate your same outputs. This is an interesting use case.

Some comments:

The documentation in the project.build module's load_inputs function must be updated to reflect the new optional character of proj_existing_builds.tab. It would be important to remind people that if they choose not to have a proj_existing_builds.tab (no initial builds), then they shouldn't have a proj_build_costs.tab file in the inputs directory or an error will be raised. 

Another option could be to add some code to let SWITCH skip loading the build_costs file if the existing_builds file is not present.

A third option could be to merge both input files into a single one, with an optional character AND having optional parameter columns (proj_overnight_cost and proj_fixed_om). This way, the user could choose to have or not existing builds and also choose to input project-specific build costs or not.

I'm not sure which alternative is best in terms of readability and efficiency. I would tend to think the last one is, since both files are indexed by the same set and are both optional (since now).

josiahjohnston:
Yeah, the load_inputs() documentation should be updated. I also just realized that the sentence about cap_limited_projects should be changed to "proj_capacity_limit_mw" and moved up a few lines.

project_build_costs.tab should still be fine if there are no existing builds. That file can specify project-specific costs for both existing and new plants. For big-picture planning studies, generic costs for each technology can be adequate, but for optimizing near-term investments, the project-specific costs would be more useful.

josiahjohnston:
Hi Mark. Can you fix the documentation and merge this in?

mseaborn:
OK, I have added a note in the docstring to say that this file is optional.

I didn't change anything about project_build_costs.tab since Josiah said it still applies to new plants.

PR #28, opened 2016-07-02
Print output variable values in a sorted way. 

bmaluenda:
This allows easier comparison of results when developing features, since printing won't be random. It also aids in visual inspection to identify odd behaviors.

This increases printing time slightly. For a 10k line output file of a variable indexed by 2 sets (i.e. dispatch for 200 projects in 50 timepoints), printing time increased from 0,63 s to 0,78 s while running in my PC.

I think the benefits outweight the increase in print time, specially so new users can have a better experience while inspecting the example output files.

### DIFF
--- a/switch_mod/utilities.py
+++ b/switch_mod/utilities.py
@@ -561,8 +561,10 @@ def _save_generic_results(instance, outdir):
             writer.writerow(['%s_%d' % (index_name, i + 1)
                              for i in xrange(var.index_set().dimen)] +
                             [var.name])
-            for key, v in var.iteritems():
-                writer.writerow(tuple(make_iterable(key)) + (v.value,))
+            # Write values sorted by indices
+            key_object_pairs = sorted(var.items())
+            for pair in key_object_pairs:
+                writer.writerow(tuple(make_iterable(pair[0])) + (pair[1].value,))
 
 
 def _save_total_cost_value(instance, outdir):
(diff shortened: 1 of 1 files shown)


switch_mod/utilities.py
```diff
@@ -561,8 +561,10 @@ def _save_generic_results(instance, outdir):
             writer.writerow(['%s_%d' % (index_name, i + 1)
                              for i in xrange(var.index_set().dimen)] +
                             [var.name])
-            for key, v in var.iteritems():
-                writer.writerow(tuple(make_iterable(key)) + (v.value,))
+            # Write values sorted by indices
+            key_object_pairs = sorted(var.items())
+            for pair in key_object_pairs:
```
josiahjohnston: Can you use:
for key, obj in key_object_pairs:
    writer.writerow(tuple(make_iterable(key)) + (obj.value,))

using pair[0] and pair[1] is less readable (in my opinion) and inconsistent with the rest of the codebase.

josiahjohnston:
Great idea. I made some suggestions for code style. 

For examples and development environment, this will totally be really useful. For some production environments, people may prefer to avoid the runtime hit, which could grow with larger datasets. Can you make this an optional parameter that defaults to false? We used this pattern in save_inputs_as_dat() with the determistic_order parameter.

bmaluenda:
Thanks for the comments, I implemented them. I'm sorry, but it seems the second commit rewrote the whole file just to modify a few lines (I'm using a new GUI on Ubuntu and it has some unexpected behaviors). If the code seems ok to you, I'll close this Pull Request and commit the changes directly so the history is cleaner.

josiahjohnston:
The code looks good to me and passes the tests. :)
..Well, it fails the 3_zone_toy_stochastic_PySP test, but that was a pre-existing issue.

That's a weird bug with your new GUI. Definitely makes it harder to review changes, and I'm glad that complete file rewrite didn't make it into the main branch.

PR #30, opened 2016-07-06
Skip example directories that lack an expectation file

josiahjohnston:
Skip example directories that lack an expectation file. This deals with the tricky issue of providing example usage of PySP, which has poor integration with python scripting (i.e. PySP lacks a python API, and instead directs you to command line tools). As we develop better integration with PySP, we will need to integrate it with automated testing, but that is too tricky to deal with for now.

For a quick fix, adding this behavior seemed less kludgy than writing phony output files that match a deterministic problem since this example is actually demonstrating how to specify and solve a stochastic problem.

### DIFF
--- a/tests/examples_test.py
+++ b/tests/examples_test.py
@@ -38,6 +38,15 @@ def find_example_dirs():
                 yield path
 
 
+def get_expectation_path(example_dir):
+    expectation_file = os.path.join(example_dir, 'outputs',
+                                        'total_cost.txt')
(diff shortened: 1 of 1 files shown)


bmaluenda:
I had locally committed a total_cost.txt file to avoid raising that error while running tests, but it makes much better sense to just skip that specific test. 
Works well!

josiahjohnston:
When I went to merge this in, I realized there was a second bug that was preventing the example directories from being included as test cases. fixed that and merged this.

PR #33, opened 2016-07-08
Added option to print output to both the terminal and a log file.

anamileva:
I added a little option to write run output to a log file (in a logs directory under each scenario directory) in addition to printing to standard output. I find this useful with large problems that take a while to solve, for diagnostics when comparing convergence behavior, etc.

### DIFF
--- a/switch_mod/solve.py
+++ b/switch_mod/solve.py
@@ -8,6 +8,7 @@ import pyomo.version
 
 from utilities import create_model, _ArgumentParser, Logging
 
+
 def main(args=None, return_model=False, return_instance=False):
 
     if args is None:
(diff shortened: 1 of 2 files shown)


mfripp:
Looks good, but I have a couple of questions.

Could the logging arguments be defined in solve.define_arguments() instead? That would avoid any need to create your own _ArgumentParser, and it would automatically add the logging options to the help system. The downside is that we couldn't do any logging before the model is created. (Arguments other than --module-list aren't parsed until after all modules are identified and loaded, so that each module can add arguments to the list and then they all get presented together in response to --help; this also enables detection of invalid arguments.) If this is a big issue, we could add a more formal "early" round of argument parsing, for things that may be needed before the modules are loaded, e.g., "--verbose" and the logging options.

Also, would this be a good time to introduce a more general-purpose logging system? At a minimum, this could generate output if the model is in "--verbose" mode, but suppress it otherwise. Or it could go further and have flag(s) indicating levels of verbosity and/or importance, i.e., some messages get output even when we're not in verbose mode; some messages get tagged as warnings, etc.

anamileva:
Yes, the separate _ArgumentParser is needed to be able to start logging before loading any modules. Having an early round of argument parsing seems like a good idea to me. The function names should reflect that, i.e. I could make the names more general than just 'logging' functions to make it clear that other early-round arguments can be included.

I think logging to a file should be a separate option from the "--verbose" option. I could see situations in which I might want to run in "verbose" mode, but not save all the output to a log file. On the other hand, perhaps selecting the log-to-file option should trigger verbose mode and/or solver-streaming?

mfripp:
Doing a formal parsing round before loading modules sounds like a good idea. You may need to take some care to make sure the pre-module-load arguments get shown along with the module arguments when the user specifies --help. This could probably be done by changing solve.add_module_args() into a more general solve.add_pre_module_args() (or similar).

My first instinct was to make --verbose and --log-run independent settings. --verbose triggers more output (if wanted), and --log-run writes a copy to a file. (By the way, for simple solves you can achieve something similar with "switch solve | tee logs/logfile.txt".) If we want different output on screen vs. in file, then this may not work, but I don't think we're there yet.

josiahjohnston:
This looks great to me, and argument parsing and help messages seem to work as good as ever. I say merge it. 

Can you bring it in by rebasing it onto master? That will give us a cleaner history. The instructions in this link explain it well. 
http://stackoverflow.com/questions/96615/git-is-it-pull-or-rebase-when-working-on-branches-with-other-people

PR #35, opened 2016-08-17
Fix docstring to refer to the correct column name, proj_capacity_limit_mw

mseaborn:


josiahjohnston:
Looks great, go for it!

PR #36, opened 2016-08-17
Make the dependency on sympy be a version-pinned dependency

mseaborn:
All the other existing dependencies are version-pinned to ensure
reproducibility, so make the recently-added dependency on sympy be
version-pinned too for the same reason and for consistency.


bmaluenda:
Agreed, go ahead!

PR #37, opened 2016-08-18
Added pip dependencies for hawaii modules

josiahjohnston:
Added pip dependencies for hawaii modules constant_elasticity_demand_system.py and r_demand_system.py. Also updated installation instructions to include dependency on R.

Matthias, does this look good to you?

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -25,6 +25,13 @@ A solver such as GLPK, Cbc https://projects.coin-or.org/Cbc or cplex.
     expensive for non-academics. It is free for registered academics who
     use it for teaching or research.
 
+R (optional)
+    R is needed for a Hawaii demand module. If you are not using that module,
+    then R is optional, but you will get an error message about rpy2
+    when installing python libraries with pip, and a second error message
+    when you execute run_tests.py from switch_mod.hawaii.r_demand_system
+    being unable to load rpy2.
+
 To use this model, either install this to a standard python library
 location or set the environment variable PYTHONPATH to include this
 directory. The latter option is probably more useful for developers. On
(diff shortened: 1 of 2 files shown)


mfripp:
Couple of minor questions/comments:

Do the tests cover hawaii.r_demand_system now? That is the only module that depends on R and rpy2. Most people probably won't use this (it's only loaded if users specifically choose it and write a demand system in R), so I'm happy to make both of those dependencies more optional. e.g., I could wrap the rpy import in r_demand_system with a try/except, and provide some advice to the users if the import fails (as well as spelling out the dependencies more explicitly in a docstring in that file).

numpy is used by r_demand_system and scuc, but again these are rarely used so we could make that an optional install, and prompt the user if they need it.

What do you think?

josiahjohnston:
That makes sense. Make them more optional and effectively exclude them from the test suite, but keep them in the main repo for convenience. As long as untested code is more of the exception than the rule, I don't see this becoming a problem.

I'm on board, but I'm not sure what the cleanest implementation would be.

mfripp:
Hmm, we could get fancy and write tests that short-circuit if the dependencies are missing (or maybe test for the correct warning message), but actually test the code if the dependencies are available. Do you know which test is trying to use r_demand_system? Or should I write one?

mfripp:
By the way, how do I edit the files you've changed or add commits to this pull request? Do I have to push changes to your repository? If you merge these changes into a new branch on switch-model/switch, then we could both work on it before merging.

josiahjohnston:
I'm fine with either the fancy approach or a simple approach of just skipping testing those files (and writing a note somewhere about how to re-enable tests). 

The relevant code is in run_tests.py. TestLoader.discover() scans the switch_mod and tests directories for .py files, and TestLoader.loadTestsFromModule() actually loads the modules and connects their doctests to the unit test framework.

I can't figure out how to hook into this without re-implementing a custom version of discover() .. discover() identifies .py files according to a wildcard pattern, then tries to load the module and scan for tests. It hits an error while loading the module (unmet dependency), and passes that on to the list of errors. 

Some other people posted this question on [stackoverflow](http://stackoverflow.com/questions/12487532/how-do-i-skip-a-whole-python-unittest-module-at-run-time) for this same use case of unmet dependencies, and failed to come up with a clean happy solution. 

Maybe Mark will have some other ideas for clean implementation? (adding him)

Good question about editing files on the pull request.. I haven't done that before. If I had committed this branch to the switch_model/switch repo, you could propose change directly. But if I push it there now, then I think I'd need to open a new pull request from that repo+branch. I don't see a way to redirect a pull request to a different repo/branch. 
I did add you as a collaborator to my personal repo/branch, so you can push suggestions there.
If you go through the github web interface to "Files Changed", you can find a button to edit the files. Clicking that button for a repo I can't write to advises me to fork the repo to my account. You can play around with that and see where that goes. I'll look for an excuse to explore that stuff myself as more pull requests come in. 

You can also browse through the pull request with a web browser and make comments on any line. I've used that effectively before in the past..

josiahjohnston:
I'm moving this to another pull request #43 that references a branch in our shared switch-model repo.

PR #38, opened 2016-08-20
Use load_aug() from switch_mod.utilities to deal with optional files

josiahjohnston:
Use load_aug() from switch_mod.utilities to deal with optional files instead of wrapping the standard DataPortal load() function with a "if os.path.isfile(): ..."

Really, just avoid using load() for any .tab file.


bmaluenda:
It's great to use a single standard input loading function. New developers won't hesitate when writing their own modules' load_inputs() functions.

josiahjohnston:
Does this look ok to commit?

bmaluenda:
It looks ready to be merged.

mseaborn:
LGTM

PR #41, opened 2016-08-22
A simple implementation of reservoir hydro

josiahjohnston:
A simple implementation of reservoir hydro that maintains minimum and average stream flow that can be specified for each time series. Also includes an example of specifying Run-of-River hydro as a variable renewable generator.

I included some thoughts about how to merge this with the more advanced Chilean hydro module in the future.


anamileva:
Looks good to me.

PR #43, opened 2016-08-22
Dependency support for experimental hawaii modules

josiahjohnston:
Hawaii modules constant_elasticity_demand_system.py and r_demand_system.py have undocumentated dependencies on R, rpy2, and numpy. Their tests fail unless you install these dependencies manually. rpy2 and numpy dependencies can be included in pip_requirements and aren't that big of deal. Installing R is more of a hassle because it requires using apt-get, homebrew, or manually downloading installation files. 

An easy fix (proposed here) is to list the extra python dependencies in pip_requirements, and to update the installation instructions to include notes on installing R.

A nicer fix would be to mark this code "experimental" in a way that will skip its tests if the dependencies aren't available.

Previous comments on this pull request are under Pull Request #37 which originated with my personal repository. I'm reposting it so the code will be in a branch of this repository where other people can use it for staging, not my personal repository.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -25,6 +25,13 @@ A solver such as GLPK, Cbc https://projects.coin-or.org/Cbc or cplex.
     expensive for non-academics. It is free for registered academics who
     use it for teaching or research.
 
+R (optional)
+    R is needed for a Hawaii demand module. If you are not using that module,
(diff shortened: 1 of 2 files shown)


mfripp:
closing this pull request because it's covered by #74.

PR #46, opened 2016-09-10
Remove non-portable documentation, prompt users to generate it for themselves.

josiahjohnston:
I started to update the documentation, but realized that the documentation that pydoc generates is not portable. If someone wants all of the links to work, they will need to generate it on their machine. Consequently, I'm improving the explanation in the README and deleting all of the documentation. An alternate method could be to have a "make" equivalent that will auto-generate the documentation when someone installs it on their machine, but I think that can wait for another day, and be an improvement on top of this. Really, we need to move to a better documentation system. I'd rather that each component's documentation be in its doc attribute, instead at the top of the define_components file. That would make it easier to read the code, and prompt developers to keep the documentation up-to-date. Doing this would mean hacking the documentation engine to examine model components and generate relevant snippets .. something that is a project unto itself.


README
```diff
@@ -7,14 +7,15 @@ modules to include.
...
-strings. This should include descriptions of each module, their
-intentions, model components they define, and what input files they
-expect. To refresh the documentation, go to the folder and run
-./make_doc.sh This documentation is not well-formatted or pretty, but it
-will do until we migrate to Sphinx and invest more effort into writing
-higher-level documentation and tutorials.
-
+To generate documentation, go to the doc folder and run ./make_doc.sh
+This will build html documenation files from python doc strings which
+will includes descriptions of each module, their intentions, model
```
mseaborn: "includes" -> "include"

README
```diff
@@ -7,14 +7,15 @@ modules to include.
...
 
-The doc folder contains auto-generated documentation from the python doc
-strings. This should include descriptions of each module, their
-intentions, model components they define, and what input files they
-expect. To refresh the documentation, go to the folder and run
-./make_doc.sh This documentation is not well-formatted or pretty, but it
-will do until we migrate to Sphinx and invest more effort into writing
-higher-level documentation and tutorials.
-
+To generate documentation, go to the doc folder and run ./make_doc.sh
```
mseaborn: Nit: end this sentence with ".".

mseaborn:
LGTM with a couple of nits.

josiahjohnston:
Like that? If not, feel free to edit this branch.

bmaluenda:
Regarding the edit, I would go for an intermediate solution, such as:

> To generate documentation, go to the doc folder and run ./make_doc.sh.
> This will build html documentation files from python doc strings which
> will include descriptions of each module, their intentions, model
> components they define, and what input files they expect.

Regarding your idea, I agree that it would be best to have the doc strings inside each component. We may be receiving a couple of new undergrad students to work with us this semester, so I'll check if one of them would be interested in implementing that documentation scheme.

Also, I've noticed that a "User's Manual" would be useful for new users that don't have a solid background in scientific computing. It could contain instructions to install all the required software (not necessarily specifying every step, but maybe redirecting to each software's online installation guide) and steps on how to run Switch models. Most of those instructions are contained in the ReadMe files, but having it all condensed in a single document could make it easier for new people to learn (I'm thinking on something like a PDF). We have 2 new M.Sc. students and a couple of undergrads who are going to work with SWITCH for some projects and I've found myself a bit overwhelmed trying to guide them through all the steps personally.
 
The other idea I came up with was working on the Switch Pyomo Workshop Josiah hosted in March and consolidating it. We could put all instructions and tutorials in that GoogleDrive folder and then just put a link to it in the ReadMe and have it be self explanatory.

The first option would be quicker, but the second one is more robust, I think. Maybe we could pool our efforts and get the second one done in less time?

josiahjohnston:
OK, I went back to the intermediate option, and pushed it into the master branch. Closing this pull request..

Benjamin, it would be fabulous if one of your undergrads could rework the documentation engine. It seems like Sphinx is the way to go, but I haven't delved into it sufficiently. Ideally, we would be able to readily mix free-written text (and diagrams) with documentation extracted from the code itself (both python doc strings, introspection of pyomo components and their doc attributes). Integration with python doc strings should be straightforward, but we would need to write glue code to do introspection on pyomo components. Eventually, Sphinx would generate the users manual (both in html & pdf). In the meanwhile, we can do a better job writing the content of the users manual, then migrate it into the Sphinx engine when that is ready.

A clear user manual would be fabulous as well; thanks for bringing it up! I'm fine with either a single "Getting Started" document or a google drive folder for that stuff in the interim. For collaborative writing, I usually find google docs is easy until formatting become important. I'm happy to help, but would you mind taking the lead and requesting help on specific pieces?

bmaluenda:
Cool, then let's begin by consolidating the Google Drive folder and work on top of that. I will try to get started on it as soon as I can.

PR #47, opened 2016-09-10
Storage implementation

josiahjohnston:
I implemented a storage module along with an example. It passes my preliminary tests, and doesn't mess up any of the other code.

This has some overlap with switch_mod.hawaii.batteries. This is a more generic implementation that can accommodate multiple storage technologies, and is more careful about energy vs power. If the duration of each timepoint is 1 hour, then using MW and MWh interchangeably (like hawaii.batteries does) is fine because it has an implicit multiplication by 1, but this will create bugs for timepoints of other durations. Another contrasting feature is the independent specification of energy and power components of a storage device. The hawaii.batteries module appears to assume no limits on power for charging or discharging, just on energy. This generic storage implementation does not include battery-specific components such as cycling limitations, or maximum discharge limits. My idea was for concrete storage technologies (like pumped hydro, batteries, or compressed air energy storage) to have custom modules under switch_mod.generators that model these additional rules as needed.

This implementation makes storage technologies an extension of typical generators, so most of the code and logic is reused. This implementation lumps charging efficiency with discharging efficiency. This formulation is valid if you interpret the energy capacity of a battery as being derated by the discharge efficiency. If you took a strict physics/engineering approach, then the actual energy stored in a battery be less than the energy you could extract from the battery, because the extraction process has some inefficiencies.

This storage module is appropriate for short-term storage that operates over the course of a timeseries (usually 1 day), but will not be able to model season storage that passes energy from one timeseries to another.


anamileva:
This looks mostly good to me. A few suggestions/questions:
1. Clarify that with this treatment the state of charge is the state of charge at the END of the timepoint (not the beginning, the middle, etc.) -- I have seen confusion about this before
2. I don't see data for either g_storage_energy_overnight_cost or proj_storage_energy_overnight_cost in the example -- no error since they are optional params, but is that on purpose?
3. I think proj_storage_energy_overnight_cost should be loaded from proj_build_costs.tab, and proj_storage_efficiency and proj_store_to_release_ratio from project_info.tab -- currently, it's the other way around

josiahjohnston:
Thanks Ana! I fixed those silly mistakes, and clarified the documentation on that point. Good to go?

anamileva:
Looks good!

josiahjohnston:
Super! I cleaned up the commit history and pushed it to the master branch.

PR #49, opened 2016-09-13
Use temporary files for handling tests of utilities.save_inputs_as_dat()

josiahjohnston:
Better handling of the temporary file complete_inputs.dat in the testing framework. Make the doctest clean up its output before finishing, and make the utilities test use a temporary directory. Also, remove complete_inputs.dat from the repository because it doesn't add anything, and it always shows up as being changed (the ordering of its contents is non-deterministic).


tests/utilities_test.py
```diff
@@ -21,13 +22,15 @@ def test_save_inputs_as_dat(self):
...
             args=["--inputs-dir", "test_dat"], return_model=True, return_instance=True
         )
-        # To do: create a temporary file name for dat_path and delete it
-        # when the test is complete.
-        dat_path = "test_dat/complete_inputs.dat"
-        utilities.save_inputs_as_dat(model, instance, save_path=dat_path)
-        reloaded_data = DataPortal(model=model)
-        reloaded_data.load(filename=dat_path)
-        compare(reloaded_data.data(), instance.DataPortal.data())
+        temp_dir = tempfile.mkdtemp()
```
mseaborn: Suggestion: it's useful to use a prefix argument, e.g.
  tempfile.mkdtemp(prefix="switch_test_")
so that if files are left behind in /tmp, it's easier to tell what's leaving them behind.

tests/utilities_test.py
```diff
@@ -4,6 +4,7 @@
 import unittest
 
 import switch_mod.utilities as utilities
+import tempfile, os, shutil
```
mseaborn: Suggestion: it can be useful to write one import per line, and to keep the imports sorted, i.e.

import os
import shutil
import tempfile

This reduces merge conflicts if two people independently add the same import (or it makes merge conflicts easier to resolve).

josiahjohnston:
How's this?

switch_mod/utilities.py
```diff
@@ -148,6 +148,9 @@ def save_inputs_as_dat(model, instance, save_path="inputs/complete_inputs.dat",
     ...     'switch_mod', 'project.no_commit', 'fuel_cost')
     >>> instance = model.load_inputs(inputs_dir='test_dat')
     >>> save_inputs_as_dat(model, instance, save_path="test_dat/complete_inputs.dat")
```
mseaborn: Wouldn't it be a bit cleaner to use a temp file here too, rather than letting it write into the source directory?

Or you could use "save_path=os.devnull".

tests/utilities_test.py
```diff
@@ -4,6 +4,9 @@
 import unittest
 
 import switch_mod.utilities as utilities
+import os
```
mseaborn: One other nit:  It's good style to keep imports of standard libraries together in one block, so you could group this with "import unittest".

switch_mod/utilities.py
josiahjohnston: I don't think its cleaner from the perspective of a code snippet that is easy to read, then copy & paste into your own project.

mseaborn:
LGTM aside from the comments above.

switch_mod/utilities.py
mseaborn: OK.

josiahjohnston:
Cool, thanks! :)

PR #50, opened 2016-09-13
Standarizing the way Switch writes output.

bmaluenda:
- Removed the save_results and _save_results functions for exclusive
use of the post_solve function. Replaced the latter in all modules
which were still using save_results. This eliminates several redundancies.
- Since most new Switch users are also new Pyomo users, there is a very
slim chance that they will be using old Pyomo versions, so I didn't import some snippet of code from save_results into post_solve which was supporting them.
Since there were no checks for solver termination conditions and only
a print statement, I put a generic printing of the condition on the
solve module, if verbose option is on.
- Added an option for sorted output of generic variables. The capacity
was already written, but no command line option was available.

Passes the tests and gives me the same output as in the examples folders. Code should be compatible with Hawaii modules, since its own save_results module has a post_solve function which will be called upon solver termination.

### DIFF
--- a/switch_mod/utilities.py
+++ b/switch_mod/utilities.py
@@ -5,7 +5,6 @@
 Utility functions for SWITCH-pyomo.
 """
 
-import csv
 import os
 import types
 import importlib
@@ -14,7 +13,6 @@ import argparse
 import __main__ as main
 from pyomo.environ import *
 import pyomo.opt
-import switch_mod.export # For ampl-tab dialect
 import datetime
 
 # This stores full names of modules that are dynamically loaded to
@@ -37,7 +35,7 @@ def create_model(module_list, args=sys.argv[1:]):
     Construct a Pyomo AbstractModel using the Switch modules or packages
     in the given list and return the model. The following utility methods
     are attached to the model as class methods to simplify their use:
-    min_data_check(), load_inputs(), save_results().
+    min_data_check(), load_inputs(), pre_solve(), post_solve().
 
     This is implemented as calling define_components() for each module
     that has that function defined, then calling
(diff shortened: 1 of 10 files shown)


josiahjohnston:
Great start! A few comments:

* When you do refactoring like this, you need to do a recursive multi-file search for your keyword. You missed save_results in several modules..
```bash
siah-laptop:switch josiah$ grep -rI save_results .
./.git/logs/HEAD:df2e1a90527e2869c6efbf329b4f72bec783482e 72493b1a2ca1c47528768ba04688b514951eeb27 josiah.johnston <josiah.johnston@gmail.com> 1471405430 -0700	rebase: added basic hawaii.save_results module
./examples/custom_extension/sunk_costs.py:save_results(model, instance, outdir)
./switch_mod/export/dump.py:def save_results(model, instance, outdir):
./switch_mod/export/example_export.py:def save_results(model, instance, outdir):
./switch_mod/utilities.py:    min_data_check(), load_inputs(), save_results().
./switch_mod/utilities.py:    # replace the old _save_results function
```
The first entry from the .git directory can obviously be ignored, but the next five hits should get fixed. 

* The "ideal" way of posting changes is to restrict changes to just one feature per commit. Adding the option for sorting the output of the generic dump is a bit off-topic. I've broken that rule a bunch of times myself, but I try to stick to it these days because it makes the code history way easier to read through. 
* Regarding that feature, you shouldn't stick that command line option into switch_mod/solve.py and use it in utilities.py. If you use a different entry point than switch_mod/solve.py and try to call instance.post_solve(), it generates an error because that option hasn't been defined. Really, the file that uses a command-line option should be the one defining it. A fairly clean way of dealing with that is to move _save_generic_results() and _save_total_cost_value() into switch_mod/export/__init__.py, add a post_solve() function that calls those two functions, add the command line option to that file, make switch_mod.export a core module listed in switch_mod/__init__.py, then finally remove references to _save_generic_results() and _save_total_cost_value() in utilities.post_solve().

Cheers,
Josiah

josiahjohnston:
I wrote that in a hurry last night, and neglected to summarize my comments into a punch-list.
My proposed punch-list (should be quick & easy)
* Do a recursive search on the entire codebase and remove the remaining references to save_results.
* Remove the sort output command line option from this pull request.
    - Post that as a separate feature, implemented in a different way (either following the export module suggestion, or by moving the command-line option from solve.py into utilities.py

Thanks again for taking on this task of cleaning up the codebase! :)

bmaluenda:
I think that your suggestion is great, so I moved the functions to the __init__.py file of the export module.
Also, cleaned up remaining references and fixed some docstrings and imports.
Finally, removed the deterministic output ordering option proposal. It is true that it corresponds to a different topic and should be pushed through another branch.

switch_mod/utilities.py
```diff
@@ -223,11 +221,9 @@ def post_solve(model, outputs_dir=None):
     # (the latter occurs when there are problems with licenses, etc)
     
     # replace the old _save_results function
```
josiahjohnston: Shouldn't this line get deleted?

josiahjohnston:
I'd suggest taking out the comment in utilities:
```python
 # replace the old _save_results function
```
Other than that, looks good to me. 👍

switch_mod/utilities.py
bmaluenda: Ok, I thought that leaving a comment like that would help when people looked for the old function, but if it isn't good practice, then I'll delete it.

bmaluenda: What do you think about the TODO comment I left? Should that be left there or should I initiate a new GitHub Issue thread regarding that topic and erase the comment lines in the code?

josiahjohnston: Hopefully any hypothetical user would also search the git logs for save_results, find this commit, and figure out what happened. I don't know, maybe Mark has another take on that?

The TODO comment looks fine to me. In my opinion, those work great for short notes, while github issues are better for longer explanations or discussions.

bmaluenda:
Thanks for the comments, merging now!

PR #51, opened 2016-09-18
Make the rhosetter's regular expression capable of dealing with exponents in the cost coefficients

josiahjohnston:
Tweak the rhosetter string matching pattern in so it isn't confused by cost coefficients that include exponents (like 2e-7, for example). This could make it break with single-letter decision variable names, but we never use those, so that shouldn't be an issue.

I tested this with the 3zone stochastic example and it works identically as before.

### DIFF
--- a/examples/3zone_toy_stochastic_PySP/rhosetter.py
+++ b/examples/3zone_toy_stochastic_PySP/rhosetter.py
@@ -52,7 +52,7 @@ def ph_rhosetter_callback(ph, scenario_tree, scenario):
     # parts because I used two sets of parenthesis. I don't care about the
     # second parenthesis that returns the indexed bits, just the larger part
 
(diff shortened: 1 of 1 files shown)


bmaluenda:
I'm not proficient with the use of regex in Python, so I can't give you a thumbs up or down right now :(

Why are you making this change? Were some cost coefficients not being recognized in some of your runs?

josiahjohnston:
It's a small enough change that I feel confident in understanding it, I just wanted to verify your stuff still works ok, especially for any examples you have that aren't in the main branch. The old behavior was looking for variables that start with a letter, followed by 0 or more letters or numbers. The new behavior uses 1 or more letters or numbers. 

I was playing around and ran into an edge case where cost coefficients were written like 1e3, or 1e-7, and the e was being mistakenly treated as a variable.

bmaluenda:
Then you should go ahead and merge this! It is a relevant edge case.

josiahjohnston:
Great! Done.

PR #52, opened 2016-10-05
Adding a new module to model complex (or not so complex) hydroelectric systems.

bmaluenda:
Adding a new module to model complex (or not so complex) hydroelectric systems.

It has a high degree of flexibility, allowing for modeling of most relevant operational phenomena.

Code can be made more efficient, but I've separated some sets and constraints on purpose so that
the module is more readable.

I pretend to upload an example sometime soon, but meanwhile you can check the module itself.

The boundary conditions for reservoir volumes are a pain, since the model will always tend to use up all available water. There are a couple of commonly used conditions:
(i) Including a water cost for the volume that is stored in the final timepoint the simulation (a Future Cost Function).
(ii) Forcing each reservoir to reach the end of the simulation with a certain volume of water.

I hard coded in the latter, though with a period resolution. So, each reservoir will end each period with the inputted final volume parameter, at least. This works well for the Chilean system, since most reservoirs only have intra-annual regulation capacity and only a couple can store water for several years.

More flexibilization needs to be added, so that people can modify the hydro system's topology within
the simulation horizon (now it's considered to be fixed) and to allow using different boundary conditions.

I tried documenting it as best as I could. If there is anything that is not clear or you find a bug, please
let me know and I'll fix it.


josiahjohnston:
Yay!

If you're going to post an example soon, I'll wait till then to review. New models are easier for me to read when I have accompanying data from a simple concrete example.

bmaluenda:
I added an example that I hope can help you examine the module. It is a uninodal system with 1 PV project and 1 CCGT, along with 3 hydro projects. There are 48 timepoints, representing a summer day and a winter day, with hourly resolution.

The hydro projects are located in 2 basins:
Hydro_Dam_1 is located right below reservoir Dam_1 and its outflows go to Sink_1.
Hydro_Dam_2 is located right below reservoir Dam_2 and its outflows go to Node_1.
At Node_1, there is a constant inflow, which is added to the outflows from Hydro_Dam_2.
Hydro_Series is located downstream from Node_1 and its outflows go to Sink_2.

If you plot the output dispatch, you should see how the dam project's tend to follow the load curve and generate a bit less during the day, when the sun is out. Natural Gas is only used at peak hours (~23 hrs).

An interesting feature of this example is how Dam_2 releases so much water during peak hours of the Summer day (hours 22 and 23), that the Hydro_Dam_2 project has to spill water, but that spillage is used by the Hydro_Series project downstream to generate up to its maximum capacity. The model prefers to do this to avoid using Gas.

There are some strange behaviors of Dam projects, with big ramps at some hours (i.e. see hours 27 and 35), where one Dam project ramps down and the other ramps up. I suspect that this is caused by multiple optimal solutions, where spending water from one dam is the same as spending water from another. But I'm not sure 100%.

bmaluenda:
I'm currently reviewing the hydro module and testing some changes. The current formulation presents numerical issues for some topologies and inflows, because of the complex problem structure that the water balance equations imply. I'm testing relaxing some of the equations by allowing spillage of water at more locations. Initial results show no more numerical issues and faster solution times.

Also, I plan to break up this module into a core hydro one and some auxiliar ones. For example, minimum ecological flows are not enforced in every region and data may not even be available (as is my case). So, I think readability will increase if I take all of those "extra" features and put them in a separate module.

josiahjohnston:
I posted some of my suggestions into pull request #61. I probably shouldn't have spent so much time merging reservoirs into water nodes, but it wasn't until I was testing that I realized reservoir consumption wasn't being accounted for in the initial implementation. Anyway, here's what I got:

Requested changes: Please do these.
* Ensure that automated testing works. hydro_system's doctests fail in run_tests.py. A simple fix is to remove that synopsis. It is now outdated and not the recommended way of loading and using a module, plus it requires adding test data to switch_mod/test_dat/. More exhaustive tests are performed by running the hydro_system example, so the doctest really isn't adding anything. I should think about stripping that out of other modules too.. 
* Enforce_Reservoir_Balance appears to have a unit conversion bug. In simplified form, it currently is:
		sum(inflows) - sum(outflows) = (final_vol - init_vol) / timepoint_duration_hrs
	Unit analysis shows that the left side resolves to m^3/s, while the right side resolves to m^3/hr. The fix is to replace timepoint_duration_hrs with (timepoint_duration_hrs*3600).
* Account for consumption at reservoirs in Enforce_Reservoir_Balance by subtracting m.res_tp_consumption
* Typo: hidraulic_location -> hydraulic_location
* Use infinity instead of 9999 for the default value of wc_capacity
* Add minimum data checks
* Remove the filtration hooks from this module (wc_is_a_filtration and the equations for FilteredFlow), and paste them into a filtration module when you implement it. This stuff was confusing on my first read, and it took me a minute to realize it was just hooks for as-yet unimplemented features.
* Use shorthand for defining cross products instead of nested for loops adding to a set
```python
mod.RESERVOIRS_BALANCE_POINTS = Set(
	dimen=2,
	initialize=lambda m: m.RESERVOIRS * m.TIMEPOINTS
```


Suggested changes: Do these if you agree. See second & third commits on pull request.
* Replace simple constraints with lower & upper bounds for variables to reduce the number of model components.
* Attempt to speed up mass-balancing at water nodes. Reduce the number of model components, and reduce the impact of nested for loops using a trick from the last code speed-up. This isn't necessary, but I wanted to play around with your code to get a feel for it, and this seemed theoretically useful. 
* Minor speedup to first_tps_in_period calculations
* Use an underscore to prefix data that is being cached in the model, like first_tps_in_period -> _first_tps_in_period
* Delete the set WATER_SINKS because it isn't used
* I like your idea of merging RESERVOIRS into WATER_NODES. I added one possible implementation as the third commit on the pull request.



Questions & Comments:
* In your stochastic programming scenario trees, have you been able to make every production cost node into a leaf node? If not, you may want to link reservoir levels across investment periods instead of constraining that externally.
* ReservoirInitialVol, ReservoirFinalVol, and DispatchWater are redundant variables. I'd consider dropping ReservoirFinalVol for simplicity, and to write Enforce_Final_Vol_Condition using ReservoirInitialVol and the flows. You could alternately write ReservoirFinalVol as an expression of the initial volume and flows.
* Have you modelled pumped hydro storage yet? 
* hydraulic_location documentation mentions cascading generation, where multiple turbines/projects are installed at the same location. The constraint Enforce_Hydro_Extraction forces all water to flow through each of the generators, which can have separate efficiencies and spill/dispatch decisions. It seems like you could also model that system in aggregate with fewer terms and identical results, unless the exact location of spilled water is somehow important. Is there a benefit to disaggregating the individual turbines?
* The non-linearities that are ignored in hydro_efficiency could be approximated via piecewise linear functions if that is a useful feature. This method is used in the fuel markets supply curve as well as the partial load heat rate curves in unit commitment. 
* I just realized that both hydro examples are production cost models without any investment decisions. I think we should either move them to the production cost directory, or add a note in their README files. 
* A few points of your documentation mention wanting to change the water network during simulation. You can't actually add elements to a graph dynamically while solving an LP, but you can specify all possible elements a-priori, and make them inactive unless a decision variable representing investments becomes non-zero. But your set of water nodes and connections won't be changing during the simulation.

josiahjohnston:
Regarding one of your earlier points: 

> There are some strange behaviors of Dam projects, with big ramps at some hours (i.e. see hours 27 and 35), where one Dam project ramps down and the other ramps up. I suspect that this is caused by multiple optimal solutions, where spending water from one dam is the same as spending water from another. But I'm not sure 100%.

Yes, that up-down ramping combo is a classical sign of multiple optimal solutions where those two dams don't have a clear loading order. If you gave the dams variable costs that were different values to prompt an implicit loading order, I expect that behavior to vanish.

josiahjohnston:
One more request:
Delete all of the outputs from the example directory except for total_cost.txt. These take up disk space, make commit logs harder to read through, and change all the time either due to rounding errors or to several equally optimal solutions.

bmaluenda:
I had forgotten to answer your questions and comments:
* I haven't linked reservoir levels across periods because that would cause the hydrologies scenario tree to explode. In the [attached image](https://github.com/switch-model/switch/files/569126/one-inv-scenario-branch.pdf) you may see the approach I'm taking. So, if only 3 different hydrologies could happen in any given year, that would imply 3^10 trajectories for 5 2-year periods. If the tree is cut at the end of each period, the number of trajectories is only 45. By analizing historical water head at reservoirs for the last 10 years in our system, it is reasonable to assume that every 1 or 2 years the water head returns to the "initial level". Behavior is quite cyclical, except for 2 reservoirs that do have inter-anual storage. In this  first attempt, I'm just going to ignore that capability. In the future, it will definitely be necessary to incorporate it and we'll have to figure out how.
*  Yes, though I had written it that way for readability. I changed my mind when I saw your formulaton, which is quite readable and reduces number of components.
* No, because I haven't had the need to (there's no pumped storage projects in Chile). It is something we should do, so I'm going to add some comments in the module for future implementation.
* I'm not sure I understand your question. If the user wants the units to be aggregated, he can just merge the several projects into one with the summed capacity. Some projects should nonetheless be modeled separately, as in the cases you mention: when they have different hydro efficiencies or you want to force independent spillage decisions (to enforce minimum ecological flows in some specific projects).
* They could, though I haven't found any research where someone actually models the efficiency as a function of the water head and/or flow. A couple of papers that focus on short term operations do it, but it is a higher level of detail than what is required for investment planning. At least in my opinion, it is not worth it, unless we start using SWITCH for more detailed production costing studies.
* I will add a note in the README files.
* You are right, I hadn't thought it through when writing the module. It is actually a perfect analogy of the transmission network.

josiahjohnston:
Looks good to me once the other merge is done.

Glad you were able to implement a reduced-form scenario tree!

Re: cascading generation for several projects at the same location
I was basically asking what the use cases were for disaggregation. 
If you have two projects at a location with two different efficiencies (a & b), you should be able to summarize them as one project with an aggregate efficiency without any data loss. So I don't understand how different project efficiencies are a good use case. 
If distinct spill decisions have relevant ecological impacts, that seems like a good use case. Do they work like that in Chile?

bmaluenda:
Regarding the cascades:
It seems you can only aggregate projects when they have similar capacities. If not, then their output power as a function of water flow becomes a piecewise linear function, instead of a simple line.

In example, take Gen A with a hydro efficiency of 2 MW/cumec and maximum output of 10 MW, and Gen B with an efficiency of 1 MW/cumec and max output of 30 MW.

If the water flow in that connection is 5 cumecs, outputs are:
Gen A: 10 MW
Gen B: 5 MW
Total: 15 MW. Efficiency in this case would be 3 MW/cumec, the sum of the individual values.

If the water flow in that connection is now 20 cumecs, outputs are:
Gen A: 10 MW
Gen B: 20 MW
Total: 30 MW. Efficiency in this case would be 1.5 MW/cumec, a linear combination of individual ones.

I couldn't think of a way of aggregating them without having to define efficiency as a piecewise curve.

josiahjohnston:
Got it. Is this a typical use case? I had assumed that cascading generators that use the same water would be sized similarly, but possibly have different efficiencies from different head heights.

One important thing I almost forgot to mention:
This model does not include transit times for water. Water sent over a connection will appear at the next node before the end of the current timepoint, rather than taking hours or days to make the journey. I don't know enough about hydro systems to understand when this is a reasonable simplification and when a different treatment would be needed. Could you add a brief discussion of transit times to the top of the hydro_systems file? Is this a reasonable concern or just something I'm making up?

bmaluenda:
At least my dataset includes cascading generators that can be different (some have twice the capacity as others). I think it depends on when they were built and by whom. Some projects are small, because they were built by medium firms (and usually spill a lot of water, because they were designed to have a high capacity factor). Other large ones were built by bigger companies which can afford to take more risk. I'm not sure that is the case, but it is a plausible explanation.
Regarding water heads, that is a real use case. A lot of cascading projects have different efficiencies because of height differences between intake and outflow. In mountainous geographies slopes can vary significantly as you move up/downstream. 

This is a great point. In my literature research I found that only in some detailed operational models are transit times taken into account. Since they add inter-temporal constraints, they complicate the formulation and increase solving times significantly. So, I ignored them for a planning perspective module. I'm not sure about their impact on investment planning decisions, given that I couldn't find any research on that. I'm also not sure about their impact on real operations, because I haven't had experience in that area.

I will add some discussion notes in the doc string with these questions. This is an initial version, so I think it is reasonable to ignore that representation for now. Though, it is good to leave notes for future implementation or at least to be aware of the limitations of the module. Great point that I had forgotten about!

PR #53, opened 2016-10-05
Add command line option to allow sorted output printing.

bmaluenda:
Default behavior is to print without sorting, to prioritize speed.

This option is extremely useful for visual inspection when developing new code or testing datasets, specially when combined with a nice diff tool.

### DIFF
--- a/examples/3zone_toy_stochastic_PySP/PySPInputGenerator.py
+++ b/examples/3zone_toy_stochastic_PySP/PySPInputGenerator.py
@@ -72,14 +72,14 @@ from pyomo.environ import *
 print "creating model for scenario input generation..."
 
 try:
-    module_fh = open(os.path.join(inputs_dir, 'modules'), 'r')
+    module_fh = open(os.path.join(inputs_dir, 'modules.txt'), 'r')
 except IOError, exc:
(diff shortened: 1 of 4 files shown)


josiahjohnston:
This feature makes sense, and the implementation in switch_mod/export/__init__.py looks fine.
I have a few comments on switch_mod.utilities.save_inputs_as_dat()
* That edit makes the line length exceed PEP8 standards. Sticking to less than 80 characters lets me pull up two files side-by-side on my small laptop screen, which makes coding and review go faster.
* If you changed the name of the function's argument, why didn't you set it to default to the command-line option's value?

bmaluenda:
- You are right, I missed that line. Broke it into two.
- I wasn't sure it would be correct to reference model options from that function, since it may be that it is being called externally just to print the dat files and I wouldn't know how to access them. I changed the PySP_Input_Generator script in the 3zone_toy_stochastic example to illustrate how I thought it would work. In my mind, the model options would be accessed when the user calls the `save_inputs_as_dat` function and if not specified, it would default to False for speed. What do you think about this?

josiahjohnston:
That sounds reasonable. This all looks good to me!

PR #54, opened 2016-10-05
Added simple timing messages for when the verbose option is on.

bmaluenda:
Should help for quick and simple profiling when developing new features.

Also helps new users understand what is happening and get a sense on the
magnitude of time that SWITCH spends in each process:

1. Creation
2. Instantiation
3. Solving
4. Post solving

### DIFF
--- a/switch_mod/solve.py
+++ b/switch_mod/solve.py
@@ -11,6 +11,7 @@ from utilities import create_model, _ArgumentParser, Logging
 
 def main(args=None, return_model=False, return_instance=False):
 
+    start_time = time.time()
     if args is None:
         # combine default arguments read from options.txt file with 
         # additional arguments specified on the command line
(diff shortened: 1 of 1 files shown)


josiahjohnston:
Looks good to me as a lightweight timing option with minimal added complexity. If/when we have a use case for sometime fancier, we can rewrite it then or switch to a different method. Note that time is displayed is wall clock time. Good job using time.time() instead of the [more ambiguous](http://pythoncentral.io/measure-time-in-python-time-time-vs-time-clock/) (but occasionally more precise) time.clock()

There doesn't seem to be a definitive recommended method for this kind of light-weight profiling. 
* [timeit](https://docs.python.org/3/library/timeit.html), is better suited for doing repeat measurements. You could use it for measuring just 1 command, but it adds some random complexity to the stack trace and isn't as versatile.
* [cProfile or profile](https://docs.python.org/3/library/profile.html) are good for a detailed look into where time is being spent, but it requires using tools like [RunSnakeRun](http://www.vrplumber.com/programming/runsnakerun/) to browse the profiling data.
* [This article](https://www.huyng.com/posts/python-performance-analysis) was a good read and turned me onto [line_profiler](https://github.com/rkern/line_profiler) and [memory_profiler](https://github.com/fabianp/memory_profiler), but it didn't point out any better standard strategies than the ad-hoc timing you added.

bmaluenda:
I used the idea that was implemented in the PySP module, because I liked the simplicity. I wanted to make this option useful for novice users, since developers should be able to use profiling tools without much trouble. The idea was to time whole processes instead of having a line by line detailed breakdown, so that:

- It should help developers know something is wrong when they change a feature and time spent in a block shoots up. Then they may choose to carefully profile `solve.py` and find bottlenecks.
- It should help new users get a sense on the resources that are being spent in each block. The increased verbosity will also help them understand how the model is prepared and run. 

I was not familiar with `time.clock()` until I read the article you mention. Would it be better to measure only time spent by the CPU working on Switch? Maybe we could output both? At a first glance, I would lean towards only using `time.time()` to keep it simple.

josiahjohnston:
Agreed. Stick with time.time() since time.clock() behaves inconsistently across platforms. Plus simple wall clock is best for newbies.

PR #55, opened 2016-10-06
Make solve.py play nice with line_profiler and memory_profiler.

josiahjohnston:
When either of those tools are used, solve.py isn't the top-level executable, so __name__ refers to the profiling module rather than switch_mod.solve, and the program crashes at a random spot because command-line option parsing breaks down.

Also, make the call to switch_mod.utilities less ambiguous so this module can be more portable and custom copies could live in arbitary directories.

### DIFF
--- a/switch_mod/solve.py
+++ b/switch_mod/solve.py
@@ -6,7 +6,7 @@ from pyomo.environ import *
 from pyomo.opt import SolverFactory, SolverStatus, TerminationCondition
 import pyomo.version
 
-from utilities import create_model, _ArgumentParser, Logging
+from switch_mod.utilities import create_model, _ArgumentParser, Logging
 
 
 def main(args=None, return_model=False, return_instance=False):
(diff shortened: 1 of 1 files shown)


bmaluenda:
Cool, I tested this with cProfile and it works smoothly.
I had met this problem when profiling and had gotten around it by profiling a very simple Python script called `profile_switch` that only had the following code:

`from switch_mod import solve`
`solve.main(['-v','--solver=gurobi','--other-options'])`

So, this fix is appreciated. I think you should go ahead and merge.

mfripp:
So I guess this means the profilers load switch_mod/solve.py as a script instead of a module? I don't see too much trouble with this change, although it sort of breaks the sense of "add myself to the module list"; e.g., we'll have to watch out if we ever rename solve.py to something else. 

Another solution might have been to profile the `switch` console command (which is also a python script) instead of solve.py. I often invoke the debugger on switch models by running `python -m pdb \`which switch\` ...`, which lets me use all the usual command line parsing and model setup, and also makes the debugger available for post-mortem inspection.

PR #56, opened 2016-10-18
Remove g_competes_for_space

josiahjohnston:
Matthias, this touches on scenario_data.py, so I want to give you a heads-up. 

This removes the unsupported and unnecessary parameter g_competes_for_space. This was intended to be used in conjunction with allowing different projects to compete for the same land area, a feature that hasn't been implemented yet. I realized that this attribute is more a function of individual projects than the generation technology, since in theory, any generation technology could compete with any other for a given tract of land. When we implement support for location competition, we don't need this generation parameter to cycle through a list of locations and associated projects. This is mostly an vestige of the AMPL implementation.

Most of the changes are to example generator_info.tab files, which is just removing that column.

Any existing scenario directories that includes that column will still run fine; extra columns are allowed and do not generate an error.


bmaluenda:
Good idea to clean up unnecessary parameters. Everything looks good to me, though I didn't check `scenario_data.py`.

PR #57, opened 2016-10-20
Add minimum uptime and downtime capability

mfripp:
I've added code to switch_mod.project.commit and switch_mod.project.dispatch to enable minimum uptime and downtime rules for generators.

To support this, I also added an indexed set `PROJ_ACTIVE_TIMEPOINTS` which lists all timepoints that each project is active. This may be useful for other purposes as well.

As an example and test, I added a minimum downtime constraint to examples/production_cost_models/unit_commit, which forces it to run the combined cycle plant at a higher level in the first timestep, which causes power to be dumped or the geothermal plant to be backed down.


switch_mod/project/dispatch.py
```diff
@@ -160,6 +160,22 @@ def init_dispatch_timepoints(m):
     mod.PROJ_DISPATCH_POINTS = Set(
         dimen=2,
         initialize=init_dispatch_timepoints)
+    # make set array showing all timepoints when each project is active
```
josiahjohnston: Move comments like this to the docstring at the top of define_components, and possibly to the doc attribute of the component itself. Ex:
```python
    mod.PROJ_ACTIVE_TIMEPOINTS = Set(mod.PROJECTS, initialize=proj_active_timepoints_rule,
    	doc=("The set of timepoints when a given project may be active. " +
    	     "This will be an empty set if a project is retired before the " +
    	     "start of the study."))
```

switch_mod/project/dispatch.py
```diff
@@ -160,6 +160,22 @@ def init_dispatch_timepoints(m):
...
+    # note: it might be cleaner to first define PROJ_ACTIVE_PERIODS, then use that
+    # to define PROJ_ACTIVE_TIMEPOINTS, PROJ_DISPATCH_POINTS, etc.
+    def proj_active_timepoints_rule(m, p):
+        if not hasattr(m, 'proj_active_timepoints_dict'):
+            d = m.proj_active_timepoints_dict = dict()
+            for p, t in m.PROJ_DISPATCH_POINTS:
```
josiahjohnston: Your iterator "p" is overwriting the function's argument of the same name. Rename this to something like "p2" and your weird pop behavior will go away.

switch_mod/project/dispatch.py
```diff
@@ -160,6 +160,22 @@ def init_dispatch_timepoints(m):
...
+                if p not in d:
+                    d[p] = []
+                d[p].append(t)
+        # note: we could use pop(p) here to reduce memory use, but for some reason pyomo does 
+        # one extra call at the start, so that doesn't work.
+        # note: we return an empty list if a project has no active timepoints, usually 
+        # for projects that retire before the study
+        return m.proj_active_timepoints_dict.get(p, [])
```
josiahjohnston: Move notes like the empty list one into the documentation at the beginning of define_components, and possibly in the doc field of the component itself. 
If you want to really clean up, you could do something like this:
```python
        # Use pop and delattr to gradually clear out and delete the cache
        active_timepoints = m.proj_active_timepoints_dict.pop(p)
        if len(m.proj_active_timepoints_dict) == 0:
        	delattr(m, 'proj_active_timepoints_dict')
        return active_timepoints
```

switch_mod/project/unitcommit/commit.py
```diff
@@ -249,6 +255,72 @@ def define_components(mod):
...
+    mod.PROJ_MIN_DOWNTIME_DISPATCH_POINTS = Set(dimen=2, initialize=lambda m: [
+        (pr, tp) 
+            for pr in m.PROJECTS if m.g_min_downtime[m.proj_gen_tech[pr]] > 0.0
+                for tp in m.PROJ_ACTIVE_TIMEPOINTS[pr] 
+    ])
+    
+    def tp_prev(m, tp, n=1):
+        # find nth previous timepoint, wrapping from start to end of day
+        return m.TS_TPS[m.tp_ts[tp]].prevw(tp, n)
+    # min_time_projects = set()
```
josiahjohnston: Remove these commented-out sections from development.

switch_mod/project/unitcommit/commit.py
```diff
@@ -129,6 +129,12 @@ def define_components(mod):
     Total_Startup_OM_Costs[t in TIMEPOINTS] is an expression for passing
     total startup O&M costs to the sys_cost module.
 
+    g_min_uptime[g] and g_min_downtime[g] show the minimum time that a 
+    generator can be committed (turned on) or uncommitted (turned off),
+    in hours. These usually reflect rules intended to limit thermal 
+    stress on generator units. They default to 0 (free to turn on or
+    off at any point) if not provided.
+
```
josiahjohnston: Also describe the new sets and constraints.

switch_mod/project/unitcommit/commit.py
```diff
@@ -249,6 +255,72 @@ def define_components(mod):
...
+    
+    def tp_prev(m, tp, n=1):
+        # find nth previous timepoint, wrapping from start to end of day
+        return m.TS_TPS[m.tp_ts[tp]].prevw(tp, n)
+    # min_time_projects = set()
+    def min_time_rule(m, pr, tp, up):
+        # if (pr, up) not in min_time_projects:
+        #     min_time_projects.add((pr, up))
+        #     print "creating min {}time constraint for {}".format('up' if up else 'down', pr)
+        # note: this enforces a simple rule: all capacity turned on in the last x hours must
```
josiahjohnston: Move notes like this into the documentation so it can be available everywhere, not just in the code. If you also want notes like this in the function itself, consider using a docstring so it can be extracted during the auto-documentation process. Ex:
```python
    def min_time_rule(m, pr, tp, up):
        """
        This enforces a simple rule: all capacity turned on ...
        """
        ...
```

switch_mod/project/unitcommit/commit.py
```diff
@@ -249,6 +255,72 @@ def define_components(mod):
...
+            # note: this rule stops one short of n_tp steps back (normal behavior of range()), because 
+            # the current timepoint is included in the duration when the capacity will be on/off.
+            if up:
+                rule = (    # online capacity >= recent startups (all recent startups are still online)
+                    m.CommitProject[pr, tp] 
+                    >= sum(m.Startup[pr, tp_prev(m, tp, i)] for i in range(1, n_tp))
+                )
+            else:
+                rule = (    # offline capacity >= recent shutdowns (all recent shutdowns are still offline)
+                    m.ProjCapacityTP[pr, tp] * m.proj_availability[pr] - m.CommitProject[pr, tp]
```
josiahjohnston: Use CommitSlackUp instead of ```m.ProjCapacityTP[pr, tp] * m.proj_availability[pr] - m.CommitProject[pr, tp]```. You forgot to account for proj_max_commit_fraction which is part of the upper bound of CommitProject, but it's included in CommitSlackUp.

bmaluenda:
We were missing this in the unit commitment, thanks!

I have only one question: why define a new Set instead of using PROJ_DISPATCH_POINTS?

The Sets that index the new constraints could be built like this:
```
mod.PROJ_MIN_UPTIME_DISPATCH_POINTS = Set(
    dimen = 2,
    initialize = mod.PROJ_DISPATCH_POINTS,
    filter = lambda m, pr, tp: m.g_min_uptime[m.proj_gen_tech[pr]] > 0.0)
```
And the downtime set is analogous.

I got the same results with this code. I think it looks a bit cleaner and defines less elements in the model. Also, I have had bad experiences with dictionaries when initializing constraints and sets, because they increase instantiation time significantly in large models. Querying dicts is usually the most time consuming activity in the instantiation of my simulations.

mfripp:
OK, I'll work on cleaning this up. But I wanted to ask, what's our policy on documenting the idea behind each part of code? I kind of prefer having it in the code itself, since it's easier to add it and find it there. I find it's hard to flip back and forth between long, abstract documentation at the top and the code itself, and the documentation at the top tends to get out of date.

bmaluenda:
We discussed with Josiah moving the docstrings at the top of each script into each of the components' `doc` attribute. This way, someone could write an auto documentation function that read all of the components and printed their attributes (using Sphinx, Doxygen, or any other tool). In addition, it would make it easier to read the modules, since documentation and comments would be located in the definition of the components themselves and there would be no need to repeatedly scroll to the top of the file. Comments regarding the code itself or specific implementation issues could go as inline Python comments right before a component is defined. What do you think about this?

mfripp:
I definitely like the idea of moving documentation near to or into the definitions of individual model attributes, since that would make it easier to write, maintain and find. However, I'm not sure Pyomo's doc attribute is the way to do this, because its main use seems to be to describe (not document) the object when you do model.pprint() or model.object.pprint() (i.e., it is not shown by help(model.object)), and because there is currently no automatic way to aggregate the doc attributes into comprehensive documentation. I'm not totally against overworking the doc attribute and/or writing custom code to parse these assignments out of our module files. But I think there may be a better way.

This [Stack Overflow discussion](http://stackoverflow.com/questions/8820276/docstring-for-variable) gives examples of creating docstrings for instance attributes (like m.PROJ_ACTIVE_TIMEPOINTS) by adding bare strings right after the assignment. The question points to [PEP 258](https://www.python.org/dev/peps/pep-0258/#attribute-docstrings), which recommends this approach, as does [PEP 224](https://www.python.org/dev/peps/pep-0224/). Some answers to that question point out that [Epydoc](http://stackoverflow.com/a/20529741/3830997) and [PyCharm](http://stackoverflow.com/a/34053178/3830997) use this method to some extent. However, all the examples of this approach seem to be limited to assignments at the top level of a module, which wouldn't help us. And both of those PEPs were rejected on the grounds that the docstring wouldn't be associated unambiguously with the particular assignment.

This [Stack Overflow discussion](http://stackoverflow.com/questions/3051241/how-to-document-class-attributes-in-python) covers similar ground, but has two answers that may be more helpful. [One](http://stackoverflow.com/a/15537029/3830997) looks like a technique that Pyomo could use to turn the `doc` argument into a proper docstring. But that's probably beyond our scope, and doesn't address the main challenge, which is generating good, aggregated documentation for the project. However, [this other answer](http://stackoverflow.com/a/9558703/3830997) says that Sphinx supports documentation of class attributes. That seems to be borne out by [this answer](http://stackoverflow.com/a/8659919/3830997) to a different Stack Overflow question. So maybe Sphinx (or similar) would be the right way to move ahead with this?

For now, I'll just leave these notes here (they could get moved into a separate "Issue" discussion), and I'll revise this particular code to follow our current documentation practices (big docstring at top of module).

josiahjohnston:
This is a great thread. :) I've introduced a summary as [issue 76](https://github.com/switch-model/switch/issues/76) and referenced this thread. Let's move future conversation over there, please.

mfripp:
@bmaluenda asked, "why define a new Set instead of using PROJ_DISPATCH_POINTS?" Here's a quick response: The main reason I built up the new sets from scratch was that filtering operations on PROJ_DISPATCH_POINTS can be quite expensive, especially if most of its members are likely to be excluded. In this case, Python has to look at every member of PROJ_DISPATCH_POINTS, and then pull out maybe 5% of them (most dispatch points are for renewable energy projects, which won't have minimum uptimes or downtimes). On the other hand, we can quickly get a list of relevant projects, and then construct a set with only the timepoints for those projects, as shown here.

I also wanted to point out that the PROJ_ACTIVE_PERIODS and PROJ_ACTIVE_TIMEPOINTS are likely to be useful in other modules (I often need them when I want to get the active timepoints for a particular subset of projects), and they in fact provide a cleaner way to define PROJ_DISPATCH_POINTS: `mod.PROJ_DISPATCH_POINTS = Set(initialize = lambda m: ((proj, tp) for proj in m.PROJECTS for tp in m. PROJ_ACTIVE_TIMEPOINTS[proj]))`. If we switched to that formulation, I think we could eliminate PROJECTS_ACTIVE_IN_TIMEPOINT, PROJECT_BUILDS_OPERATIONAL_PERIODS and PROJECT_PERIOD_ONLINE_BUILD_YRS. That's on my short-term to-do list, but it should be in a different branch than this one.

PR #58, opened 2016-10-20
minor fixes to solve and export modules

mfripp:
I  tightened the patch on `Expression.construct()` so it only applies to Pyomo 4.2 and 4.3. The corresponding bug has been fixed in Pyomo 4.4. 

I also changed string formatting to use `.format()` instead of `%` throughout solve.py. `%` should not be used in new code, since it will not work with Python 3.

I also changed the _save_generic_results() code to allow non-indexed (single-value) variables, which exist in some of my models.

### DIFF
--- a/switch_mod/export/__init__.py
+++ b/switch_mod/export/__init__.py
@@ -78,20 +78,25 @@ def _save_generic_results(instance, outdir, sorted_output):
         if not isinstance(var, Var):
             continue
 
-        index_name = var.index_set().name
         output_file = os.path.join(outdir, '%s.tab' % var.name)
         with open(output_file, 'wb') as fh:
             writer = csv.writer(fh, dialect='ampl-tab')
-            # Write column headings
-            writer.writerow(['%s_%d' % (index_name, i + 1)
-                             for i in xrange(var.index_set().dimen)] +
-                            [var.name])
-            # Results are saved in a random order by default for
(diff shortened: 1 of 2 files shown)


bmaluenda:
This looks great.

Sorry for the Python 2 string formatting, I'm going to stick to the compatible format from now on.

mfripp:
I've been using a very old version of switch with a different patch to `Expression.construct()` for my main iterated demand-response model. When I started using it with the current version of SWITCH this week, I discovered that the patch to `Expression.construct()` in `solve.py` is actually being applied too late, i.e., after the model is constructed. Just shows, I should be writing tests for every bit of new code. Now I've moved it so the patch is applied before the model is constructed. 

I also removed the deprecation warnings, because this patch is part of our "official" ongoing support for Pyomo 4.2 and 4.3. Eventually, when we're ready to abandon Pyomo 4.2 and 4.3, we could add a deprecation warning for a while and then remove this patch.

josiahjohnston:
Looks good to me. 

I haven't ran into the Pyomo expression issues yet, so I can't test that fix against positive and negative controls, all I can say is that the code passes all of the existing tests in my environment. 

Good addition to _save_generic_results(). If having single-value variables scattered across files gets to be a pain, you could toss them all into a single file with two columns (name, value).

PR #64, opened 2016-11-03
Fix to the min-build bugs from pull request #63.

josiahjohnston:
* Removed dangling reference to project.min_build
* Moved peak coincident load from local_td to load_zones so it could be available to more modules.
* Use system-wide non-coincident peak load as a basis for binary constraints on capacity. Using an overly large number like cumulative system-wide energy from the entire study horizon was producing errors in how glpk handled the constraint that enforces a binary decision.

### DIFF
--- a/examples/discrete_and_min_build/inputs/modules.txt
+++ b/examples/discrete_and_min_build/inputs/modules.txt
@@ -1,5 +1,4 @@
 local_td
 project.discrete_build
-project.min_build
 project.no_commit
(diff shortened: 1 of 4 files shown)


bmaluenda:
GLPK's behavior is really weird. At least it's good that we found the bug now so a new user won't run into it.

The example runs smoothly and all tests pass. The new upper bound looks reasonable to me. I say go ahead and merge.

PR #65, opened 2016-12-18
Add install instructions for ArchLinux, update pip requirements.

mateoatr:
- Update the pip required packages
- Add install info. for ArchLinux.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -52,6 +52,30 @@ pip install --user -r pip_requirements.txt
 # Install glpk & glpsol
 sudo yum install -y glpk glpk-utils
 
+--------------------------------
+Instructions on ArchLinux
+
+# Check python version is set to 2.7.x. If you are currently using Python 3 or higher you can set a virtual environment using virtualenv
+python --version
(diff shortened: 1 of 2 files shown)


bmaluenda:
Including rpy2 and psycopg2 in the `pip_requirements.txt` file will raise an error if the user doesn't have R and Postgresql installed. Since both packages are used only in the Hawaii modules, I'm not sure we should include them in the requirements list. That would force users to install both programs (even if they are not going to use the Hawaii modules) or to install all other packages manually.
I can think of 2 workarounds:

- Removing them from the `pip_requirements.txt` file and adding a note in the installation instructions file so that users that want to use the Hawaii module can install them manually.
- Moving them to a new `pip_requirements_hawaii.txt` (or any other adequately named) file.

That is if we consider that users will not necessarily have or want to have both R & Postgresql installed in their machines.

mateoatr:
@bmaluenda Ok, this makes sense. For the moment I'll just leave the ArchLinux instructions then.

PR #66, opened 2016-12-18
More minor fixes

mfripp:
These commits do the following:

- rename `m.proj_end_year` to `m.proj_final_period` to clarify which year it refers to (also slightly improve the calculation, e.g., using `m.period_start + m.period_length_years` instead of `m.period_end`, because `m.period_end` could refer either to the last full year of the period or the moment when the period ends (the start of the next period)
- provide post-mortem debugging, activated via a `--debug` flag
- allow single-point heat rate curves  in `project.unitcommit.fuel_use` (e.g., a plant that always produces exactly 8 MW if it is committed); a few minor bug fixes and refactorings in this code
- remove `m.DispatchProjByFuel` from `project.no_commit` module because it cannot be generalized neatly for models that use `project.unitcommit`. Note: this term has been replaced by code in the `hawaii.rps` module that uses binary variables to distinguish between RPS-eligible and RPS-ineligible fuels (and artificially prevents use of multiple fuels in the same project in the same period). That code could eventually be generalized to allocate production to individual fuels and moved back into the core code. However, doing that would introduce more binary variables than the RPS-oriented code, and would artificially limit fuel-mixing. This term is only useful when there's an RPS in place, so I'm leaving it in `hawaii.rps` for now.

These changes should not interfere with any existing code (all tests pass).

### DIFF
--- a/switch_mod/generators/storage.py
+++ b/switch_mod/generators/storage.py
@@ -164,7 +164,7 @@ def define_components(mod):
                 m.Storage_Build_Summation_dict[p2] = set(
                     (proj, bld_yr)
                     for bld_yr in m.G_NEW_BUILD_YEARS[m.proj_gen_tech[proj]]
-                    if bld_yr <= p2 <= m.proj_end_year[proj, bld_yr])
+                    if bld_yr <= p2 <= m.proj_final_period[proj, bld_yr])
         # Use pop to free memory
         relevant_proj_build_years = m.Storage_Build_Summation_dict.pop(p)
         return sum(m.BuildStorageEnergyMWh[proj, bld_yr] *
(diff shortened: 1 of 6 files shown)


josiahjohnston:
Hi Matthias & everyone.
What do you think about replacing period_end with period_length_years in the input files? period_end is only used to calculate period_length_years and is subject to multiple interpretations. Dropping period_end would simplify the model and remove ambiguity and guesswork in the code.

mfripp:
I would be fine with that -- we always need to know the period length anyway to set the end year. How about adding this capability, but still supporting the end_year approach, with deprecation warnings, for a few weeks or months? That would smooth out the transition a bit.

PR #72, opened 2017-01-21
Various minor fixes

mfripp:
This code does a few things:

- adds an "--interact" flag which allows the user to interact with the model after it is solved
- corrects the standard version number (no '.' before 'b')
- fixes some bugs in the iteration counting code for multi-level iteration

All tests pass, so I plan to merge it immediately.


bmaluenda:
The interaction option is a great addition!

Issue #70, opened 2017-01-08
Standardize installation and execution


mfripp:
I think we need to standardize how SWITCH is installed and run, because we seem to have two different versions going on. I'll describe below how I do it, which is my recommended approach. But I would like to hear if people have other suggestions. I would then like to rewrite the instructions so users can get up and running pretty easily.

I install SWITCH by cloning the repository, then cd'ing into the `switch` directory and running `python setup.py develop` or `python setup.py develop --user`. This installs the package in-place, so I can edit it and use it at the same time (and if I want to change to another installation, I just go there and run `python setup.py develop` again). Users who don't want to edit the package can instead run `python setup.py` or `python setup.py --user`. Once I put the package on pypi and conda, they can just do `pip install switch` or `conda install -c switch-model switch` .

These commands work well under anaconda, and would probably work with homebrew python. With the standard system python, users might need to use `sudo` with the system-wide versions.

These commands do a few useful things:

- install the `switch_mod` package in the python system path
- install dependencies specified in setup.py (Pyomo and its dependencies for now, but the list could be extended)
- install the `switch` command-line script (equivalent to `python -m switch_mod.main`)

I then setup models in various locations in my file system (not inside the `switch` folder) and solve them by running `switch solve` or `switch solve-scenarios`. If I edit the local copy of the switch package or use `git pull`, those changes are automatically reflected the next time I use the command-line script. This works well, and makes it easy to setup other users and share models with them (via separate repositories for each model).

As I understand it from the INSTALL file, other SWITCH users are editing their `PYTHONPATH` to point to the switch repository and then using `pip install --user -r pip_requirements.txt` to install the dependencies. I don't know how other users are activating the `switch` command-line script, if they use it at all. This approach has a few disadvantages compared to my approach, which make it difficult to give general installation instructions for new users: 

- it is hard to write general instructions for users to modify their PYTHONPATH appropriately; 
- using pip requires installing pip on some systems, and may cause problems for people who use anaconda
- it's nearly impossible to write cross-platform instructions to install the `switch` command-line script, which is  very handy, especially for new users

All of these problems are addressed automatically by setuptools, which is used by setup.py. This takes care of all the cross-platform issues other than installing the solver.

Any objections to making `setup.py` the standard way to install switch, via the various commands listed in the second paragraph? Or making `switch solve ...` the standard way to run it?

rodrigomha:
Hi everyone,

[...]

PR #74, opened 2017-01-24
Use setuptools and switch command line tool

josiahjohnston:
Updated installation instructions to use setuptools, and usage instructions to use the switch command line script.
Added dependencies and extras to setup.py, set minimal Pyomo version to the glpk bugfix, updated entry_points syntax, implemented pointers from https://packaging.python.org/distributing/
Renamed pip_requirements.txt to requirements.txt to reflect standard python naming conventions.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -5,19 +5,7 @@ Python v2.7
     to do so. We have verified our code works with v2.7.9 and 2.7.10. We
     had some difficulties with our testing framework on v2.7.3.
 
-Pyomo
-    To install Pyomo, you can run this command:
-
-        pip install --user -r pip_requirements.txt
-
-    The pip_requirements.txt file records the versions of Pyomo and other
-    Python libraries that Switch is tested against.
-
-    Alternatively, you can see Pyomo's instructions for installing Pyomo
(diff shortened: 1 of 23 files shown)


mfripp:
I will merge the changes to setup.py from pull request #75 into here. However, there's one question we should address -- should the "extras" be defined by capability (demand_response, run_tests) or by type of dependency (r, psql). I'm inclined to use the former, since it will make more sense to the user. Let me know if you have any concerns about that.

mfripp:
I've now merged my version of setup.py with Josiah's. To keep things simple for the users, but also allow a lightweight "standard" installation, I put all the optional dependencies in an "advanced" extra. I have also moved my "readme.txt" code from issue #70 into INSTALL and README. This makes both of those files a lot more verbose, but that might be a good thing.

For now, I'm assuming people should install using the --editable command, since that's the most plausible way to work if they are downloading the repository from github. Later, we can have a "git" option with --editable and a "pip install switch" option for people who don't want to see the source code and examples.

josiahjohnston:
Looks good to me.

PR #75, opened 2017-01-24
Make most external packages optional

mfripp:
This changes the hawaii demand response modules so they don't load any external packages (numpy, scipy, rpy2) unless the demand response modules are used in a model. It also changes hawaii.scenario_data to not load the psycopg2 package until it is needed. This could become standard practice for specialized modules with external dependencies: don't load the dependency unless the module is actually used to construct a model. This allows the unit tests to run successfully even if the extra packages are not installed.

To formalize this, setup.py now lists Pyomo as the only required package, and lists numpy, scipy and rpy2 as dependencies for the `demand_response` "extra" (see [http://setuptools.readthedocs.io/en/latest/setuptools.html#id14](http://setuptools.readthedocs.io/en/latest/setuptools.html#id14)). It also lists testfixtures as a requirement for the `run_tests` extra. People who want to use these extra features can install the required packages by running this command after installing the switch package: `pip install switch[run_tests,demand_response]`. Or they can install switch and the extras at the same time by running something like this in the switch directory: `pip install --editable .[run_tests,demand_response] --user`.

setup.py also now lists `testfixtures` in the `tests_require` variable, and defines a `test` command which calls run_tests.py. This allows users to execute `python setup.py test` immediately after downloading the repository.

If people run hawaii.scenario_data.write_tables() without the psycopg2 package installed, they will be instructed to install psycopg2 separately. This seems like a better solution than including it as an extra dependency, because only core developers of the hawaii model will use this package (to access our back-end database). If external users need it, we could add psycopg2 as a dependency for a `hawaii_database` or `database_access` extra.

This pull request also includes a bug fix for switch_mod.solve_scenarios when running on Windows.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,7 @@ required = [
 # then use 'pip install --user switch[run_tests,demand_response]' to install optional dependencies
 extras = {
     'run_tests': ['testfixtures'],
-    'demand_response': ['numpy', 'scipy']
+    'demand_response': ['numpy', 'scipy', 'rpy2']
 }
 
 packages = ['switch_mod']
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,7 @@
 """Setup script for SWITCH. 
 Use "python setup.py install" to install a copy in the site packages directory.
 Use "python setup.py develop" to activate SWITCH while preserving its current location.
+
 """
 
 import os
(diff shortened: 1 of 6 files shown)


mfripp:
Closing pull request because this branch has been merged into the standardize_setup branch (pull request #74).

josiahjohnston:
Looks good to me!

PR #78, opened 2017-01-28
Code restructuring and input upgrade scripts

josiahjohnston:
* Restructured modules into investment and operations directories, to replicate
the formulations found in literature for expansion planning.
* Phased out the gen_tech module, so all parameters must now be specified on a project basis.
This matches scientific literature models and makes code easier to understand.
* Restructured financial calculations into a financials package.
* Add support for project build decisions that have been determined outside of the optimization
* Make modules.txt explicitly list all python modules that it wants by their full names. 
* Input upgrade script (and updates to example directories)
* Renamed fuels.py script into energy_sources.py


mfripp:
What do you think of renaming the "operations" sub-package to "operation"? That would be consistent with "investment" (and with the SQL notion of using singular names for tables, although that's not really relevant here).

switch_mod/upgrade/upgrade_2_0_0b1.py
```diff
@@ -0,0 +1,274 @@
...
+
+    # Does 'modules' need to get renamed to 'modules.txt'?
+    modules_path_old = os.path.join(inputs_dir, 'modules')
+    modules_path = os.path.join(inputs_dir, 'modules.txt')
+    if os.path.isfile(modules_path_old):
+        shutil.move(modules_path_old, modules_path)
+
+    ###
+    # Upgrade module listings
+    with open(modules_path, 'rb') as f:
```
mfripp: This should use the same logic and possibly the same arguments as solve.py to find the modules file (and maybe the inputs dir). The best way might be to share command line arguments with solve.py. This could be similar to how solve_scenarios.py calls solve.get_option_file_args().

josiahjohnston: Makes sense. I'll look into merging those snippets of code.

josiahjohnston:
I hear people say "operations" a lot as a term of art .. Operations Research, Operations Management, etc. I don't remember hearing anyone use the singular "operation" in those contexts.

mfripp:
I'm looking more at the parallel between the words investment and operation. Both of these are [nominalizations](https://en.wikipedia.org/wiki/Nominalization) of verbs, so you can say "investment in the power system" and "operation of the power system", but "operations of the power system" is more awkward and more general, not something specific you can optimize. When talking about power plants, we talk about the cost of "operation and maintenance", not "operations and maintenance." I think this folder holds modules that pertain to operation of the power system, rather than a grab-bag of things that are in the general realm of operations.

bmaluenda:
Same here. I'm more used to reading and hearing "operations", but either option would make the contents of the directory clear enough for users.

mfripp:
Both "investment" and "operation" can be either singular or plural. To me, the singular form seems to have a closer tie to the verb and therefore to a specific action (operation of equipment, investment in equipment). The plural forms seem to encompass a wider range of activities, including things outside the scope of SWITCH ("operations" includes power plants and billing, "investments" include power plants and shell companies).  We wouldn't claim to be optimizing the "operations" of any individual power plant, but we are optimizing its "operation" (choosing when to operate it). "Operation" is also commonly used in the power system, e.g., for "operation and maintenance expenses". So if you don't have a strong preference, I think it would be more consistent to write them both as singular. I can do the work for this, but I don't want to steamroll your opinions.

josiahjohnston:
Go for it if you're motivated.

josiahjohnston:
The notes on our discussion of new features are in [this google doc](https://docs.google.com/document/d/1QjVpc17xxs6T6IGfai9PI0kfbsqOdprrWWbXLVSWY _w/edit#).

The goals of reorganizing the packages were:
* Make it easier for newcomers to find relevant parts of the model
* Reduce the amount of cognitive overhead for 
* Continue enabling alternate implementations or levels of complexity (eg. no_commit/commit

What about renaming "financials" into "economics"? I think the idea was

mfripp:
SWITCH inherently has crosscutting classifications: timescale (build, commit, operate) and technology (projects, transmission, local_td, EVs, etc.). When placing the code into a directory tree structure, we inevitably have to decide which classification to use first and which one second. 

We can make it clear in papers and documentation that the model has separate classes of variables and constraints for each timescale. But I don't think that means we have to use timescale as the top-level division for our code, especially if it interferes with the modularity of the code.

This seems to be the most relevant part of the discussion document with respect to dividing the model by timescale (emphasis added):

> **The expansion planning problem is regarded as a bi level optimization problem where the investments (construction) in Gx and Tx “are decided first” and then operations are simulated to check their cost. This structure could be replicated in the module directories.** We could have a directory for construction decisions (project_build and trans_build) and a directory for operations (transmission and generation, where trans_dispatch would be located inside the former (and renamed to transport_model_flow), and no_commit and unitcommit in the latter). 
> We see the following benefits arising from this:
> Replicates the mathematical modeling structure present in literature: construction & operations

However, the other benefits reported for the new directory structure actually come from partitioning the code by technology (transmission, projects, etc.) rather than timescale:

> User-devs will find it easier to add modules in those categories. For example, someone could add a dc_flow module inside the transmission subdirectory, inside the operations directory, parallel to transport_model_flow. Another case would be if someone added a new Unit Commitment formulation, which would go in the generation subdirectory inside the operations directory, parallel to unitcommit. 

An important aspect of this version of SWITCH is that it is easy to add, remove or replace individual technologies. The proposed directory structure seeks to provide these benefits by partitioning the code by technology _after_ partitioning by timescale. But the benefit will be greater if we divide by technology in the first place, and then divide by timescale when relevant (e.g., proj_dispatch.py and proj_commit.py within a projects package). If, as currently planned, we split every module into fragments that can be placed into the investment and operation directories, it will actually make it harder for new users to understand where to find the modules related to a particular technology or how to replace them; e.g., separate dc_flow modules will have to be created to replace trans_build and trans_dispatch, which are widely separated in the file system.

Given that we have to make two distinctions in the model (timescale and technology), I think it makes more sense to divide the code primarily by technology, and secondarily by timescale. In papers, we can start by discussing the timescales, possibly even presenting the relevant project and transmission modules in that order. Then we could move on to self-contained discussions of other technologies, covering both construction and operation decisions together. I'm reluctant to choose a much less convenient structure for the code just so we can say, "look, it's divided between construction and operation at the _top_ level".

What do you think of this?

josiahjohnston:
I understand that reasoning. I'd like to get Benjamin and/or Rodrigo in a conversation before moving forward. They are on vacation, so we might need to table this for now.

@bmaluenda and Rodrigo, if either of you see this and can do a conference call this week (maybe Tuesday), please send me an email. Otherwise, Matthias and I will work on other things.

Matthias, if you have any other ideas for restructuring code, can you post them into the [google doc](https://docs.google.com/document/d/1QjVpc17xxs6T6IGfai9PI0kfbsqOdprrWWbXLVSWY_w/edit#) where we've been discussing that stuff? That should make it easier to take notes during our next conference call. Thanks!

mfripp:
Here are some responses to the third comment above, from Josiah.

As I noted above, I think the objectives related to code discovery, learning and replacement would be better served by partitioning the code primarily by technology, rather than timescale. If we take this approach, I would recommend guidelines like this: 

1. simple technologies can be modeled via a single file in switch_mod
2. conceptually related technologies can be grouped into a subdirectory or tree below switch_mod
3. if a technology is too complex to represent in one file, it should be split into separate files, possibly in their own directory; often these will be split into build/operate files

This leaves some ambiguity about whether you will find related technologies or separate parts of a single technology in a subdirectory (trans_build, trans_dispatch, security_constrained_power_flow and local_td might all be in a transmission subdirectory). But that can always be resolved by refactoring (e.g., split transmission directory into trans & dist subdirectories, possibly bundle trans_build and trans_dispatch into a single subdirectory). The key point is, however this is done, people will always know where to look for a particular component, and they will find all the files for that component in one place.

Dividing the code by timescales would help to conceptually distinguish "build" variables from "operate" variables. But I don't think it would actually give us an automatic way to select all the "build" variables to use in decomposition methods like Progressive Hedging. However, as you noted, there are other ways to do that. In the past, I have kept an explicit list of "build" variables, but that could become more difficult as more different things get built. One way to simplify this would be to adopt an "everything is a project" approach, so that the only build variables are "ProjBuild". Another way would be to register "build" variables in a list, as we do with energy production and consumption, and will eventually do with reserves.

> What about renaming "financials" into "economics"?
> The only physical constraint I see in fuel_markets is the supply curve; is that what you're talking about? If we called this package "economics", I think it would be fair to say that physical constraints on scarse resources could fall under "economics".

The fuel_markets_expansion module is similar in detail to the proj_build module, i.e., it defines specific physical projects that could be built, which would expand the fuel supply curve in each region. I think these projects and fuel supplies in general are pretty far afield from discount rates, interest rates and amortization. For now, the financials module could just be a single file. Eventually, it may get more complex, when we offer utility-style depreciation or other amortization options.

Hopefully we'll hear back from Benjamin and Rodrigo soon. Meanwhile, I'll put specific recommendations for reorganization into the shared Google document.

rodrigomha:
Hi guys:

So our main idea with Benjamin with these changes was to order the code for
newcomers. Having everything in switch_mod was really confusing and we
think that it was a better option. The order we decide it is not definitive
and definitely can be modified to improve it.

So one question, when you say ordering for technologies, did you mean
having a directory for every type of generator? In electrical engineering,
when someone says for example generation technologies, he is referring to
their energy source. In such case, that would mean that we need a directory
for every source, like wind, coal, gas, etc. There is no need to separate
everything in different directories. I think I'm confused on how this would
work, so

From what I understand, dividing by timescales it is what we (me with
Benjamin) proposed. We want to have the build part, that is performed in a
period scale, in investment, and everything else in operations, that is
performed in timepoint scale. The problem as you mentioned, is that
separate build with dispatch (for a specific technology) in different
directories.

Having the code ordered by technologies it is a good idea and it could
work. But in my opinion, I would create a directory about transmission,
that would have trans_build (that creates the lines) and we can have
transport_model, or dc_flow, or what we want to dispatch our network. For
such purpose, creating the set of directed lines should be in the
transport_model, and not in the build, since that set will be used for the
transport_model and probably we would use other technique to decide how to
calculate the flows in a dc model.
Then I would only have a directory about generation (or project), and
inside of it we decide the proj_build and proj_dispatch. And then inside of
it, we can have another directory or how we want to dispatch our plants
(for example using no commit, or ucommit or maybe other future techniques
like convexhull commit). Probably it makes sense to include all special
types of generation in other directories inside of our generation (or
project) directory. This need to include the storage, hydro, and other
things (maybe DR?) inside of it.

I think we can discuss this after I arrive to Santiago. Right now I'm in
Peru, Arequipa in a hotel with crappy wifi, so probably I can't make the
call from here. I will be arriving in Santiago at February 6, so after that
monday I would like to have a Skype session to decide what would be the
best solution to have everything ordered and in a simple structure.

Regards

mfripp:
Sorry, I should have defined what I meant by "technologies".  This is more like "role in the power system", rather than individual generation technologies. Right now I'm thinking the top-level directories would be something like this:

- projects
- transmission
- balancing
- energy_supply
- policies
- reporting
- timescales.py
- financials.py
- upgrade

There's a complete tree in the [Google doc](https://docs.google.com/document/d/1QjVpc17xxs6T6IGfai9PI0kfbsqOdprrWWbXLVSWY_w/edit). I think this would meet everyone's clustering goals, would have a natural home for everything (no need to punt on things that don't fit), and would keep related components close together and easy to find.

bmaluenda:
I ran into a bug in the upgrade script when updating the Chile input dataset from v2.0.0b0 and I'm not sure on what is causing it.

New generator technology build cost values are not properly read in, so these cannot act as default values for projects that don't have specific overnight and fixed_o_m costs. I found that commenting lines 318 and 319 in the upgrade script removed the bug. Seems that when the `lz_cost_multipliers` parameter is absent from the load_zones.tab file, NaN values are used instead of 1.0, resulting in NaN values for the costs when these are multiplied.

I'm leaving it commented for now, but if you took take a quick look it would be great.

josiahjohnston:
This is strange.. Lines 227-230 should be setting NaNs to 1 where they exist, and setting lz_cost_multipliers to 1 if the column wasn't in the input files. 
Can you post the Chile input dataset so I can replicate the problem?

josiahjohnston:
Per our earlier phone conversation, we're going to table the dataset upgrade issue for now, and revisit it after you post a dataset.

PR #80, opened 2017-02-22
Adding module for demand shifting at no cost.

bmaluenda:
Allows assessing the value of load shifting.
Maximum demand curtailment and increments are defined per load zone and timepoint.


josiahjohnston:
Would you mind taking everything out of the outputs directory except the objective function value in your example? Those files change non-deterministically between runs and clog up git diffs.

bmaluenda:
We assumed that that convention was being broken for the copperplate0 example to provide output files for the most basic example. We mistakenly maintained that behavior on this new module, but since it is not basic, it shouldn't have output files. We'll remove them now.

josiahjohnston:
I posted some proposed changes. What do you think? 

It's mostly about avoiding confusion between **Load Shifting**, an emerging class of DR that we we are modeling, and **Load Shedding**, the traditional class of DR that we are _not_ modeling. I tried to pull vocabulary inspiration from this DR report: http://www.cpuc.ca.gov/General.aspx?id=10622

Also, what do you think of renaming the example directory to dr_simple? In the README you can explain that this adds DR to the copperplate0 example.

I was thinking about renaming this to load_shifting_dr_simple, but that seems like a mouthful.

bmaluenda:
Where did you post them? We can't find them :(

josiahjohnston:
Sorry, I forgot to push it.

bmaluenda:
Nice changes, and the total cost remains the same.
We are merging!

PR #82, opened 2017-03-06
Updated RPS module to base rps target on sales (not generation)

josiahjohnston:
Updated RPS module to base rps target on sales (not generation) to reflect RPS policies in most US states (including California and Hawaii). This ignores transmission and storage losses and reflects known bugs in most existing RPS legislation.

**References:**
[Hawaii RPS Summary](http://programs.dsireusa.org/system/program/detail/606)
[California RPS Summary](http://programs.dsireusa.org/system/program/detail/840)


bmaluenda:
It works the same way in Chile and it makes sense to reflect actual policies, even if they present flaws. Additionally, this formulation is going to speed up model resolution.
I say merge.

PR #83, opened 2017-03-07
Simplify the formulation by removing a set that was only used in one module

josiahjohnston:
This implementation uses Constraint.Skip instead of defining a custom set

### DIFF
--- a/switch_mod/energy_sources/fuel_costs/simple.py
+++ b/switch_mod/energy_sources/fuel_costs/simple.py
@@ -54,14 +54,13 @@ def define_components(mod):
         within=PositiveReals)
     mod.min_data_check('FUEL_AVAILABILITY', 'fuel_cost')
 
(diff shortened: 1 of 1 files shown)


bmaluenda:
Looks good, though I think its best to leave the `proj` or `pr` index, until we rename the components.

**Related comment**: This type of change reduces model components and, thus, simplifies the formulation. Nonetheless, model instantiation time is increased, because of the `if` condition in component construction. I have found that it is usually faster to build additional sets to avoid those conditional clauses.

Maybe the solution is to reduce components to have a simpler and easier-to-grasp model, and to eventually publish a "fast" version of SWITCH that avoids conditional clauses and other speed hacks. That level of speed is not needed for most users, I think, so we could focus for now in building a user-friendly program.

josiahjohnston:
Leaving the set in for speeding up things seems reasonable. I didn't realize this would have a negative impact on that. 
I don't think this one set will have a large impact on user comprehension.

PR #84, opened 2017-03-07
Streamline dispatch sets

josiahjohnston:
Replaced PROJ_ACTIVE_PERIODS with PROJ_ACTIVE_TIMEPOINTS_IN_PERIOD to simplify end-uses in rps_simple. Replaced the internal ACTIVE_PROJ_PERIODS set with a private function.

### DIFF
--- a/switch_mod/generators/core/dispatch.py
+++ b/switch_mod/generators/core/dispatch.py
@@ -141,15 +141,10 @@ def define_components(mod):
 
     """
 
-    def _ACTIVE_PROJ_PERIODS(m):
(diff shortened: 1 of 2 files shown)


mfripp:
Looks fine to me, but the names don't seem to match our final pattern yet. Will "PROJ_ACTIVE_TIMEPOINTS_IN_PERIOD" be renamed later to "ACTIVE_TPS_FOR_GEN_PERIOD" (or similar)?

josiahjohnston:
Yes

PR #85, opened 2017-03-08
Local td patch

josiahjohnston:
Update local_td to correct minor bugs and enable reasonable treatment of Distributed Energy Resources (DER), accurately accounting for distribution losses, by creating a distribution energy balance node for each load zone that is connected to the zone's central energy balance node via a local_td pathway that has a certain loss rate. 

local_td used to have an irregular treatment of the distribution loss term. It multiplied the distribution loss rate by load rather than by power withdrawn from the grid (and sent over the lines), which underestimated distribution losses by about 5% in the examples.
Also, the prior treatment of T&D capacity was rated on power received rather than power sent, so it underinvested in local_td by around 5%.

local_td now creates a new distributed energy balancing node (or “virtual bus”) that is connected to the zone’s central bus through a local_td pathway that has a fixed loss rate. Things that impact the distributed energy node may register with its energy balancing list for proper crediting of distribution losses. This implementation prohibits the distributed energy node from delivering energy back to the main grid because my first implementation of two-way flow modeled after transmission lines with two non-negative flow variables resulted in a large new loophole for dumping energy (that is probably an open bug for Tx), my second implementation idea of a single unsigned flow variable would not have enabled accurate calculation of line losses, and we don't have any active use cases of distributed PV sending energy back to the grid on a zonal level (plus existing infrastructure probably wouldn't tolerate reverse power flows on a zonal level).

I updated the examples’s objective function values to match the correction, and verified that when I disabled those two bug fixes that the results were identical, indicating that the discrepancies were due to those bugs fixes and not due to unexpected bugs from rewriting the code.

I also added an optional “proj_is_distributed” flag to generators.core.build and its input files that will allow distributed PV to be credited with reducing local T&D requirements. When the local_td module is included, it will move distributed generation from the load zone’s central node energy balancing equation to the distribution node’s balancing equation.

I retained the prior behavior of local_td that reinterprets lz_demand_mw. load zones interprets lz_demand_mw as withdrawals from the main grid that include distribution losses, whereas local_td has reinterpreted it as energy delivered to customers that don’t include distribution losses. I don’t love this behavior, but I wanted to limit my changes in this commit.

I also updated copperplate1’s inputs to demonstrate local_td behaviors of investing in distributed PV to avoid local_td infrastructure investments. This required updating the solar prices to make solar on the margin, but the new values are actually reasonable, given that SunShot goals of $1/W have already been achieved according to a GreenTechMedia report.

I defined a new set (ZONE_TIMEPOINTS = LOAD_ZONES * TIMEPOINTS) for indexing various components because each ad-hoc indexing (e.g. Param(m.LOAD_ZONES, m.TIMEPOINTS, …) would create a new implicit indexing set for each component that was visible in the model dump. This probably isn’t a big deal for execution speed or memory footprint, but would prohibit us from automatically merging components that shared a common index, which was an idea we had for simplifying export.


rodrigomha:
This looks really good. Now is really clear where the power is being consumed.

Also, I agree that is not perfect to have different treatments for lz_demand depending if local_td is being used or not. However, this a good approach to solve the issue that we want the demand to be exact input that users define in the input file if they don't use the local_td module.

The constraint of Meet_Local_TD is working with the DR module? We are still forcing that the investment has to be greater than the maximum peak parameter (that could be less if the DR module is being considered). For me it's fine, however I still prefer to mention that DR may specify a more complex constraint.

My opinion is to merge it if we add that commentary about DR module and a more complex constraint.

josiahjohnston:
I updated the code to make the DR module interact with local_td, fixed an issue with peak demand constraints, and updated the dr_simple example to illustrate these dynamics.

I converted lz_peak_demand_mw[ZONE*PERIOD] into lz_expected_coincident_peak_demand[z,p] which is indexed on a sparse set and will not default to max(lz_demand_mw) when data is absent to prevent the initial static load from overshadowing DER's impact on peak shaving.

I also fixed a minor bug that was neglecting to account for distribution losses when forcing m.LocalTDCapacity  to meet lz_expected_coincident_peak_demand.

These two bug fixes changed the objective function values in a few examples, so I updated those as well.

bmaluenda:
Sorry for the late comment, but nice work! Just reviewed it and everything seems ok. I think that the use of Distributed Nodes is elegant and allows modeling all phenomena of interest. It is great that you found a way for both distributed generators and demand response to lower the local T&D requirements.

PR #86, opened 2017-03-10
Planning reserves

josiahjohnston:
A simple implementation of capacity reserves margins which have been staple of the energy industry for decades. This definition of capacity-based reserves is inadequate for resource adequacy studies if energy-limited generators such as hydro or storage play a significant role in the system, and other formulations will be required to fully address resource adequacy requirements.

The basic constraint is:
```
(Capacity - Load) / Load >= capacity_reserve_margin (default 0.15)
```

Each generator has a timeseries ```capacity_value[g,t]``` which defaults to capacity factor for variable generators and 1 otherwise. It can be overridden in an input file. 

**If local_td is excluded:**
```
Capacity[z,t] = sum(capacity_value[g,t] * GenCapacityTP[g,t]
                    for g in zone if not storage[g]) 
  + sum(net_storage_dispatch[g,t] for g in zone if storage[g])
  + LZ_TXNet[z,t]
Demand[z,t] = lz_demand_mw[z,t]
```

**If local_td is included:**
```
Capacity[z,t] = sum(capacity_value[g,t] * GenCapacityTP[g,t]
                    for g in zone if not storage[g] and not distributed[g]) 
  + sum(net_storage_dispatch[g,t] for g in zone if storage[g])
  + LZ_TXNet[z,t]
Demand[z,t] = WithdrawFromCentralGrid[z,t]
```

Note, reserve capacity is implemented as a dynamic list ```CAPACITY_FOR_RESERVES```, so other modules can register qualifying capacity.


rodrigomha:
This looks really great. This is what usually makes sense in this type of capacity reserves. Also great job in including that references to validate the model. Finally, I think standardizing all lists as dynamic components will be helpful for new users to be aware what is a list and what is not.

Just a few things: I can do a commit this Monday to fix a typo in the README file of the dr_simple example. (two times solar). Also I think that I can add a comment to the constraint Enforce_Planning_Reserve_Margin as:  "Enforce_Planning_Reserve_Margin[z, t] is a constraint that sets that the sum of all the reserves capabilities is greater than the capacity requirements" like in the others modules (see load_zones for example).

This also opens again the question if we want to include information about the components only in the doc string or a brief description before defining the components as it is right now. In my opinion, we should keep the brief description as it is right now, in order to focus on the paper.

bmaluenda:
I have a couple of questions:

1) If the local TD module is not included, distributed generators won't provide reserve capacity. However, you consider non-distributed renewables as sources of capacity through their `proj_max_capacity_factor` parameter. Was this a behavior you intended?

2) Enforcing the capacity margin per load zone leads you to include storage dispatch, demand response, and transmission imports. This creates complicated constraints that mix investment and operational decisions and could significantly slow down solution in larger problems. Did the inclusion of these capacity sources appear as a recommended practice in the references you reviewed?
Additionally, it seems to me that this could induce weird behaviors in the model, because the line between capacity reserves and operational reserves begins to blur. Operations would be carried out in a way that 15% idle capacity is available at all times, overlapping with quickstart or spinning requirements. Maybe I'm looking at this the wrong way?

rodrigomha:
Just to answer a few questions of how this is done in Chile:

1.- It's usual to not consider distributed generation to provide reserve capacity. This is because it's complicated to know specifically if they will be available when it's needed.

2.- As we know, in usual models, we use a system-wide reserves per period instead of doing per zones and timepoints. I agree that including this constraint could slow down our solution times, but I'm not sure if that would happen.
In my opinion I like this module, since is a complete formulation of capacity reserves, that give us the flexibility to do it the way we want. If we are afraid of solution times, we could implement a simpler version of it (e.g. capacity_reserves_simple), that could be system-wide, per period instead of timepoints, ignores distributed generation, and uses the average capacity factor of renewable projects (averaging all the capacity factors in a period to obtain the average_capacity_factor).

2.- Also, I think you're right, this could create conflicts with quickstart or spinning requirements. 

What do you think guys?

josiahjohnston:
Thanks for reviewing this and providing insightful comments.

Benjamin 1. I suspect you are talking about line 115. Is this correct? 
That is a convoluted if statement that is difficult to read. I think you misread it; distributed resources are counted when local_td is excluded. To avoid confusion, I'm rewroting that to be easier to read. 

Benjamin 2. 

Agreed; these capacity reserve margin constraints are like simplified Satisfy_Load constraints. Yes, including scheduled imports, distributed generation, demand response, and efficiency programs is standard for California. Well, scheduled imports only count if they are contracted "firm" imports (not for "interruptible" imports). California has requirements for local resource adequacy, in additional to system-wide constraints. My inclusion of storage in there was trying to be consistent with transmission rules and a little conservative. California now has enough clarity on energy duration requirements for capacity reserves (3 hours, by the way), that we could encode something more
complex for storage contributions than just expected energy, but I didn't want to get that deep for this module, and this approach seemed defensible as an approximation.
I agree that the line between capacity reserves and operating reserves can be blurry, and the trend with resource adequacy requirements makes those lines increasingly blurry.

I don't understand what you mean by, "Operations would be carried out in a way that 15% idle capacity is available at all times, overlapping with quickstart or spinning requirements." Capacity that qualifies for capacity reserves can also be used towards other energy or reserve products; there is no requirement that the capacity be idle, just that it is connected to the system. In contrast, operational reserves require the generators to be in a certain state: capacity allocated as quickstart needs to be idle, you can't double-book a generator to operate at full capacity and provide spinning reserves, etc.

Usually the planning reserve margin is applied to either annual or seasonal peak demand timepoints, so we could simplify this constraint to operate on fewer timepoints per year. 

I can also define planning reserve areas that contain one or more load zones over which capacity reserve margins must be satisfied. 

Rodrigo:
Agreed that it can be complicated to tell if a distributed resource will be available for planning reserve capacity. In California, there is a whole process for assessing whether a resource is qualified to provide capacity reserves. Ultimately, that ends up as a list of qualifying facilities, so I can add a gen_qualifies_for_capacity_reserves parameter.

josiahjohnston:
I implemented and tested those changes. Does this look ok to you two?

rodrigomha:
For me seems OK. I think we should merge it with our v2.0.

Also I really like the example, since it has a system reserve and a specific load zone reserve, and the first one uses all timepoints, and the other one only the peak load. So a new user can understand quickly how to construct the planning reserves data.

I agree that how to use this module should be documented in detail, in particular in the part of selecting all_timepoints or peak_load. I like the flexibility of the module. Nice work on adding the capability to select for each project if they are qualified or not to provide planning reserves.

PR #87, opened 2017-03-22
Renaming components

bmaluenda:
Applied the renaming procedure! Summary of changes can be found in the commits.
All tests pass and everything looks good at a glance. A more detailed inspection is necessary to polish documentation references or small components I may have missed. Should be good enough to move things forward.
Let me know what you think.


rodrigomha:
I'm still checking all the files, but I realized that we still have this gen_tech parameter in all examples that right now I think is useless. Do you guys agree that we should remove the gen_tech column move the column of gen_energy_source to the second one? I can do that commit if you agree with it.

bmaluenda:
I think the gen_tech parameter can still be useful for two cases:

- Enables processing output according to generation technology and not only by energy source. in example you could export the energy produced by CCGT gas plants separately from that produced by OCGT gas plants, or maybe separate diesel OCGTs and combustion engines.

- Leaving the door open for customization. In example, a custom module could assign a default value to ramp rates on a technology basis.

It is fairly "cheap" to get the information regarding technology on each generation project the user is going to input. That parameter is usually the easiest to find, so I don't think that removing the component from the formulation is going to reduce any noticeable burden from the user.

josiahjohnston:
Benjamin, great articulation of the major points!
This mostly enables aggregation of results, and also provides a hook for specifying default values across a given subset of generation projects. 

Rodrigo, I agree that this parameter is currently virtually useless. Our export functionality in the main codebase is essentially absent. Exporting results is super-important thought, and I definitely want to provide hooks for future export/pandas code that spits out relevant summaries. 

-Josiah

PR #89, opened 2017-04-01
Operating reserves

josiahjohnston:
I generalized and improved the spinning reserve module from the Hawaii code. This makes extensive use of dynamic lists so other modules can register as either providing or requiring spinning reserves.


rodrigomha:
Nice work Josiah.  You managed to include both NREL 3+5 or Hawaii requirements in the model. It looks really good for the Up and Down requirements. Great work in including the greater contingency as an up requirement. 

My comments are the following ones:

1) We should mention that this module is for using it with the commit module. I know that is mentioned in the dependencies (commit.operate is listed as a dependency) but I think it will also be good idea mentioning it in the notes.
2) There are some typos with the work contigency and contigencies, instead of contingency and contingencies (missing an n).

Rodrigo

bmaluenda:
Great work in providing a flexible framework so regional modules can hook into a spinning reserves module!

Everything looks good, except for one question: Wouldn't the default value of 2 of the `contigency_safety_factor` force the system to be providing reserves in excess?

I understand that this aims to prevent the largest unit to provide reserves for itself, but the system will end up doing so regardless of the safety factor. That parameter will only require that an equal amount of reserves is provided in addition. Furthermore, what would prevent the model to allocate the whole of the reserves to the largest unit? It will probably be the cheapest.

One easy fix around this is applying what is done by our ISO: determining a static reserve requirement based on the largest possible contingency of the system (usually the outage of the largest generator unit), which has to be satisfied in every hour. The unit that would provoke the largest contingency is then simply excluded from the provision of reserves in all hours (usually large units are not equipped with frequency regulation controllers). This, of course, has its drawbacks, such as an over-conservatism of the requirement in several hours.

Another question just popped into my mind: Does this module limit the amount of reserves each particular unit can provide? Some units, such as hydro turbines, can provide almost all of their turndown ratio as reserves, because of their high ramping rates. On the other hand, thermal units can only provide a fraction of their turndown ratio, since their technical parameters limit them. So, instead of the boolean parameter `gen_can_provide_spinning_reserves`, a PositiveReal parameter could be specified with the amount of spinning reserves in MW that the unit can provide.

What do you think about these points?

josiahjohnston:
@rodrigomha Thanks for the catch! I added a bit to the documentation and corrected the misspelling. At least I was consistent in my misspelling ;)

@bmaluenda Good questions.
**Issue 1: `contigency_safety_factor` overestimates requirements**
The trick you describe is better suited to production cost modeling because it won't work if you don't know in advance which generation project will be creating the largest contingency. I considered adapting that trick, but that would require having one set of contingency constraints for each generation project, since any project could potentially be the largest on the system.. Something like:
```
sum(up_reserves[g,t] for g in GENS if g != g_contingency) >= CommitGen[g_contingency,t]
for (g_contingency, t) in GENS * TIMEPOINTS
```

My initial thinking with `contingency_safety_factor = 2` was planning for worst-case scenario where the largest committed unit was devoting all of its capacity to upward reserves. In virtual every case, this will be an overestimate, especially since with current grids, the largest generators are usually baseload and don't provide any reserves. 

I committed a variant of the spinning_reserves example that illustrates this problem. In the highest-stress hour of this system (timepoint=3), the largest committed unit is a combined cycle mid-merit gas plant. This mid-merit plant is committing most of its capacity and only producing energy, while the single-cycle peaker gas plant is committing a matching amount of capacity devoted entirely to upward reserves to deal with the other plants. This is feasible with contingency_safety_factor set to 1, but will be infeasible for most higher values. 

I've come up with a way of doing a tighter upper bound on contingency requirements:
```max(ProjectContigency[g,t]) + max(CommitGenSpinningReservesUp[g,t])```
It will still overestimate, but to a lesser degree. I can't figure out how to implement an exact solution without having an inordinate amount of constraints. 
What do you think? estimate+simplify or exact+complexity?

I guess I could try to implement both of them and let the user decide which to use..

**Issue 2: Upper and lower bounds for spinning reserves**
The version you reviewed was using DispatchSlackUp and DispatchSlackDown for the amount of upward and downward spinning reserves available. These slack variables take into account minimum run requirements for thermal plants as well as committed capacity (the slack vars are defined in the unit commitment package). The hydro_simple module doesn't register minimum run requirements with unit commitment (instead directly constraining DispatchGen), so those min run requirements won't be considered. To force those to be considered, we'll need to convert CommitLowerLimit from an expression to a variable, then have hydro_simple constrain CommitLowerLimit in addition to DispatchGen. I imagine a solution to hydro_network would be similar.

I tweaked the formulation to use variables `CommitGenSpinningReservesUp[g,t]` and `CommitGenSpinningReservesDown[g,t]` instead of `DispatchSlackUp` and `DispatchSlackUp`. The new variables are constrained by the slack expressions, but can take lower values. This deals with a potential issue with discrete unit commitment, so contingency requirements can be  DispatchGen + CommitGenSpinningReservesUp instead of CommitGen.

josiahjohnston:
Ok, so I'll leave the contingency overestimation option of `contingency_safety_factor` in place for this pull request, and we can come back to it later.

My idea for dealing with issue 2 would be explicitly describing ramping speeds of generators, and for reserve requirements to explicitly describe their required response times. In the degenerate case, we could have one class of spinning reserve requirements (as implemented now), but in the general case, we would have several classes of reserve requirements with various timescales and magnitudes, and CommitGenSpinningReservesUp would be indexed by generator, reserve product, and timepoint. The model would then add an additional constraint on CommitGenSpinningReservesUp/Down for a given reserve product based on the generator's ramping speed and the reserve product's required response time.

I think this is similar to what you suggested, except the calculation of maximum ramp rates would be endogenous instead of exogenous. Having the ramp rates endogenous would also give us more flexibility in evaluating alternative formulations of reserve requirements, which seems like a useful feature since so many utilities are rethinking reserve requirements as they integrate more renewables. 

One colleague tried using Switch for production cost simulations, but he went for a different tool when he realized we didn't consider ramping speeds which was highly relevant for his analysis on impacts of imperfect forecasts. Ramping limits is an important thing to model, but I would prefer to table it for the moment if possible. Would you mind if we postpone ramp rate considerations for a subsequent pull request?

bmaluenda:
Sounds absolutely reasonable. Specially since it would probably involve writing a new module with high detail to be used for production costing.

PR #91, opened 2017-04-16
Added feature for interactively exploring outputs from a previous model run. 

bmaluenda:
I've often needed to explore results exported out of a previous model solution and found that using the structures of the Switch model itself make it a lot easier, specially to experiment with new export functions. So, I enabled a feature to load variable values from a previous run into a new instance and return an interactive instance to the user. This saves considerable time than just running the model in interactive mode, which requires re-solving the instance (costly for large datasets).

To implement this in a clean way, I changed the index of existing transmission capacity to "1000", instead of "'Legacy'". This may conflict with users using relative years instead of absolute ones. Should I just set it to -1?

Also, added minor fix to the upgrade script.


josiahjohnston:
I like the idea of this feature. I posted some suggestions. My implementation of reading in variables may be slower, but allows for 'Legacy' .. mixing numeric and non-numeric values in a single column.

Hard-coding it to 1000 seems weird and random (reminds me of Fortran style). I guess we could use 0 or -1 instead of 'Legacy'; I don't know why but that seems cleaner to me. I'd prefer not to use a numeric constant because it is less communicative than words 'Legacy' or 'Existing'. I do see how using a numeric constant simplifies comparison operators, but it could create other problems as you point out.

I would like to rename the command line option from explore_solution though. Maybe something like "reload_prior_solution" or "reuse_prior_solution_and_export"? "explore_solution" suggests that it will dump you into interactive mode which is not the case.

I fixed the legacy transmission line issue here, but we should port that to the master branch separately to have a cleaner history.

switch_mod/solve.py
```diff
@@ -526,28 +559,18 @@ def solve(model):
...
-    elif (results.solver.termination_condition == TerminationCondition.infeasible):
+    if results.solver.termination_condition == pyomo.opt.TerminationCondition.infeasible:
         if hasattr(model, "iis"):
             print "Model was infeasible; irreducible infeasible set (IIS) returned by solver:"
             print "\n".join(c.cname() for c in model.iis)
         else:
             print "Model was infeasible; if the solver can generate an irreducible infeasible set,"
             print "more information may be available by calling this script with --suffixes iis ..."
         raise RuntimeError("Infeasible model")
-    else:
```
bmaluenda: I have no idea why this section got deleted. I may have done it accidentally. Reverting it.

bmaluenda:
Reverted several unwanted changes and changes applied in other branches.
Renamed the command line option.
Will resolve conflicts and merge.

bmaluenda:
I was initially reluctant to implement a check for float or string type data in every iteration of the for loop when loading a solution file because of increased computation time, but I agree that it may be worse to have outputs with poor legibility. Specially since this feature will be rarely used and time and resources spent loading an instance will still be much less than what would be needed to re-solve an instance (so it is still a great feature to add).

GitHub was claiming that there were conflicts that had to be resolved manually in the command line before merge. I just squashed and merged this branch into master locally and pushed that to the master branch at GitHub, and never got any conflict warning. It seems GitHub was confused by the directory's renaming. So, I'm just going to reference this Pull Request in the master branch's commit message and close this branch.

josiahjohnston:
I can revert the copperplate0 output files and `switch solve --help` bugfix if you want to keep this branch clean. I was tired and trying to finish so I could eat supper when I pushed stuff last night. I figured sharing something sloppy with you quickly was better than postponing indefinitely till I got it really clean. Sorry for the mess..

bmaluenda:
The new copperplate0 outputs are such a small and harmless update, that I don't see a problem in leaving it in this branch! I say don't worry and just leave it here.

PR #93, opened 2017-04-17
Add small Hawaii (Oahu) test case

mfripp:
This adds a small Hawaii model to the standard set of examples and tests. This gives fairly good test coverage of the Hawaii modules, and illustrates use of some interesting not-quite-standardized capabilities that have been implemented in that package. (Some of those capabilities have been moved over into the core SWITCH code, and the Oahu model will eventually be updated to use the standard features instead of modules from the switch_mod.hawaii package.)

This also updates the testing logic to change the working directory to the relevant example directory before running each test. This enables use of files that are expected to be in the current working directory when a model is run (options.txt, iterate.txt, potentially scenarios.txt). It also eliminates the need to specify an inputs directory for each test, since the default logic will now work.


josiahjohnston:
For the record, we discussed this in a conference call in Mid-April and asked for the datasets to be condensed to they could be ran in a fraction of a second as part of the testing framework. Tiny datasets also are easier to manually walk through when you are trying to debug. 
(I don't think disk size for inputs was a concern, although we try to keep larger files out of the git repo history. )
We asked Matthias to reduce the number of distinct renewable projects and/or timepoints to simplify the example. Progress on this pull request has appropriately received lower priority than finishing the switch-pyomo paper.

PR #94, opened 2017-04-17
prepare switch_model package for upload to pypi

mfripp:
This is a pretty minor pull request -- it just renames the pypi package to switch_model and changes the authors to "Switch Authors <authors@switch-model.org>". I plan to merge it immediately, since it shouldn't affect any other functionality.


josiahjohnston:
LGTM

bmaluenda:
We agreed that we should eventually rename the `switch_mod` directory to `switch_model` for consistency. I was thinking that that could actually take little time and we could do it now. Should I add a commit with that change? Or is it better to go a bit slower with these changes?

josiahjohnston:
I also thought that might be a fairly easy change. I say go for it if you have time.

rodrigomha:
As Benjamin said, I think it would be nice to modify the directory to switch_model, at least before we submit the paper. If you have some time to do it Benjamin, let's go.

bmaluenda:
Done. In order for the `switch` command to work, I just reinstalled the package by running `pip install --upgrade --editable .` after the modifications were made.

mfripp:
I didn't see the renaming of swich_mod to switch_model until after I posted 2.0.0b1 on pypi. That is now installable via `pip install switch_model`. So I have now incremented the version number to 2.0.0b2. I also fixed the input-directory upgrade code so it will rename modules from switch_mod.* (or omitted) to switch_model.*. I will post 2.0.0b2 to pypi after this pull request is merged.

bmaluenda:
How would someone with 2.0.0b1 version inputs upgrade its dataset to version 2.0.0b2? From what I understand, this code would only upgrade from version 2.0.0b0. For now, I think this shouldn't be a problem (we all have 2.0.0b0 inputs), but we should be clear on how we are going to write upgrade scripts for future versions (so people's input sets don't get stuck in-between).

josiahjohnston:
Sounds like this branch and pull request turned out to be more complicated than initially planned.. It may have been better to stage the switch_mod -> switch_model package renaming in a separate branch.

I cleaned up the upgrade scripts so that everything works cleanly. This implements a clean modular pattern of incremental upgrades that supports a spectrum of input versions. This is the pattern I was hoping to achieve during the massive renaming update, and I'm glad to have done it now so that future upgrades will hopefully be easier and smoother. 

I started to do this as a pull request onto this branch, but that was starting to fork off before the tip which would have complicated merging, so I decided to join in the free-for-all and overwrite the tip. I hope nobody minds.

PR #95, opened 2017-04-18
Instructions for using virtual environments

josiahjohnston:
For the moment, I’m treating these as advanced installation instructions for developers and power-users, but eventually it might be nice to integrate these into the standard installation instructions.

I have personally found virtual environments to be a huge time- and sanity-saver, and am hoping these instructions will demystify and simplify the use of virtual environments by others.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 .DS_Store
 gurobi.log
 switch_model.egg-info/
+venv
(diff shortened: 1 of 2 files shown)


PatyHidalgo:
I've been using the virtual environment in linux and it was worked nicely. It could be helpful for other users to learn how install and use it (specially if they are in a host where they cannot install packages globally).
Thumbs up!

gschivley:
I'm checking out the repository and model, thanks for making it all public! Any reason to use virtual env rather than conda environments? I only ask because the instructions already talk about using an anaconda prompt.

mfripp:
I think Josiah got into virtual environments about the same time I was switching over to conda, so you'll see mentions of both. Either one should work fine.

josiahjohnston:
I'm closing this PR because these changes got incorporated into the master branch some time ago with other commits.

Issue #92, opened 2017-04-16
Sunk costs


bmaluenda:
I just noticed that **transmission sunk costs** -the capital and fixed O&M costs of existing capacity- are not considered in the objective function. This is due to the `TX_BUILDS_IN_PERIOD` set not including builds indexed by 'Legacy'.
This is not a problem by itself, but it happens that the model does include **generation sunk costs** -the capital and fixed O&M costs of predetermined capacity-. If the model allowed for early decommissioning or mothballing, these costs would not be sunk, but we still haven't enabled those features.

So, for consistency, either Tx sunk costs should be included in the OF or Gen sunk costs should be excluded. Either way, both total costs and (total - sunk) costs should be reported on model exit.

josiahjohnston:
Good catch. I posted a fix to your pull request #91, but we should port that to the master branch separately since it is a distinct bug fix, mostly unrelated to your new feature in that pull request.

josiahjohnston:
Posted fix in pull request #96

PR #96, opened 2017-04-19
Miscellaneous bug fixes and minor features

josiahjohnston:
- Existing transmission lines now have their sunk costs carried into the objective function. Update example expectations accordingly. Issue #92 
- Ensure that executing `switch solve --help` outside of a project directory will not cause a crash.
- Update outputs of the copperplate0 example (the only example that includes a full set of outputs).
- Have setup.py use the version number from the switch_model package
- Implement `switch --version` for getting the version number from the command-line tool.


josiahjohnston:
Good catch. I did some digging and changed that to a more robust method similar to what you suggest. Method 3 at https://packaging.python.org/single_source_version/

josiahjohnston:
Can I get a yay or nay so I can close out this branch?

mfripp:
It looks mostly OK. However, I am concerned that you dropped the idea of having a band of capacity that does not participate in the minimum downtime constraint (previously calculated as `committable_fraction`). Without that term, the model can turn off some capacity, and then get around the min-downtime rule by turning on other capacity which is actually forced off by gen_max_commit_fraction, e.g., due to a  maintenance outage.

josiahjohnston:
Ok, I moved the min up/down time futzing to a separate branch and am going to close this one. I was concerned about leaving the transmission sunk costs bug up in the tagged version of the repo for so long. Matthias, can you either push this patched version to PyPi or give me access/show me the ropes so I can do it?

PR #102, opened 2018-02-14
General-purpose improvements from the WECC branch...

josiahjohnston:
More exports, utility function upgrades like automatically pickling the an entire solution object to disk and reloading it when requested, optionally plotting results, updates to a few examples to make them consistent with the current code.

The one potentially controversial thing is renaming "dispatch.txt" to "dispatch-wide.txt". This was an attempt to avoid confusion with the new dispatch.csv file that has detailed dispatch information in a "Tidy data" normalized form.


bmaluenda:
These improvements look nice, especially the standardization of the dispatch output. If tests pass you could merge.

josiahjohnston:
The tests had been working in my old virtual environment, but when I cleared that and did a fresh install I ran into issues with the min_data_requirements doctest, and a funny pandas warning message in the examples/storage case. The doctest bug has been showing up on and off for months and seem to be caused by pyomo changing their print statement in BuildChecks. It was unrelated to the rest of this pull request, but it needed to get fixed. I ported those doctests to the unittest framework and made them more robust. I also tweaked the "no emissions" value in dispatch export from None to zero to address the pandas warning that occurred with the storage example from that column being entirely empty.

PR #103, opened 2018-02-15
Update to 2.0.0b4

mfripp:
This branch includes various upgrades to support the case study on obtaining reserves from batteries or demand response. In particular, it adds the spinning_reserves_advanced module, which allows definition of multiple spinning reserve products (e.g., regulation or contingency).

There are also a number of minor bug fixes and feature improvements, as shown in the commit log. 

All tests pass.


bmaluenda:
All the minor fixes and improvements look good to me, though I haven't read the Hawaii package in enough detail as to be able to check those other commits.

josiahjohnston:
The spinning_reserves_advanced module needs an example to demonstrate its intended functionality and test it. Should we dust off #93 (Small Hawaii Test Case), or would you like to add a new one to exercise the different types of reserve products? 

In the interim, I made a spinning_reserves_advanced example that was should be equivalent to the spinning_reserves example, but using the new module. It has some kind of bug that yields a different total_cost. Not sure if that's a bug in the basic or advanced version, or if it's an expected behavior that I didn't anticipate. Can you take a look?

If spinning_reserves_advanced can do everything in spinning_reserves (plus more), would you be ok having this replace spinning_reserves? I'll want to make sure they are equivalent first, but I like the structure of the advanced version, and don't see a need to keep two modules that duplicate functionality. 

I just committed some edits: a new example, updates to documentation, some bug squishes, removing references to undefined components, and similar clean-ups.

josiahjohnston:
Oops, just noticed that my push didn't go through on Friday like I expected because of a glitch on my end. Just finished pushing.

mfripp:
I would certainly be happy for spinning_reserves_advanced to become the only spinning_reserves model. Just didn't want to push it in case anyone objected.

I am traveling for the next few weeks, but will look for a chance to review your commit and test. We could also add a test case for the more advanced abilities -- will have to look for a chance for that too!

mfripp:
@josiahjohnston I think this is about ready to merge to master. I'm also thinking of bumping the version number to 2.0.0. Seems like as good a time as any. Let me know if you have any objections. 

Can we live with just the simple `spinning_reserves_advanced` example that you made for now? It now matches the older version, which is nice. (Although we may now want to drop the old version...) I can add a better test case later, once I'm through this big summer paper push.

PR #115, opened 2019-06-21
Switch 2.0.6

mfripp:
(Note: most of this comment and Josiah's response have been moved to issue #129 for future implementation.)

This is a place to discuss the next release of Switch. These are my immediate goals:

- simplify dependencies
  - drop dependencies where easy (sympy, numpy, testfixtures)
  - group most remaining dependencies into one additional set? (`conda install switch_model_extras` / `pip install switch_model[extras]`)
  - require users to explicitly install some specialized dependencies ad hoc? (rpy2 and numpy for iterative demand response, psycopg2-binary or eventually google-cloud-bigquery for accessing the Hawaii back-end database)
  - should we include testing-oriented packages (pint, maybe testfixtures?) in extras, or just list them as something to add ad hoc if running tests? (may be moot if we can figure out a way to get our tests to run without needing pint, which pyomo needs during testing but does not include in the distribution)
- streamline installation instructions
  - recommend this sequence for most users?
    1. `conda install -c defaults -c conda-forge switch_model` or `pip install switch_model` (+find glpk somewhere) for most users?
    - these users can look at the source code on github or in their system install directory if needed, which is how I interact with Pyomo
  - recommend this sequence only for people who want to edit the source code and contribute back?
    1. `git clone https://github.com/switch-model/switch.git && cd switch`
    2. if running anaconda: `conda install --only-deps switch_model`
    3. `pip install --editable .` or `python setup.py develop` (note: `conda develop` doesn't install command-line scripts or dependencies; see https://github.com/conda/conda-build/issues/1992#issuecomment-322588270)
- maybe add some commands to give easier access to source code hidden in a site-packages directory:
  - `switch find <module>`: report file path to specified module (possibly just a submodule within switch_model)
    - then users can view or edit source code via commands like ``atom `switch find switch_model` ``, ``mate `switch find pyomo` `` or  maybe ``atom `switch find discrete_commit` ``.
  - `switch install examples`: copy examples directory to local directory
  - `switch solve --trace [[<module>|<file>[:<function>|<line>]], [<module>[:<function>]], ...]`:  invoke the debugger (a) when particular callbacks are called, (b) when any callback in the specified module is called (if no function specified), or (c) whenever any callback is called (if no modules specified). 
    - This could be implemented either via tests when we call callbacks, or by creating tracing wrappers for the specified callbacks, or by subclassing Pdb and setting breakpoints in the right places (that is nice and general but may run slower and ties us to Pdb?)


mfripp:
@rodrigomha and @josiahjohnston I added a commit with simplified code for the rhosetter in the PySP example, with no dependency on `sympy`. I _think_ my code should work, but it seems like the underlying implementation was probably not working already. e.g., when I run the command below, I get a message that it has data for gen_max_capacity_factor, but the key for the data is not in the indexing set VARIABLE_GEN_TPS_RAW.

Since this is your code, would you be able to see what you can do to get it to work? 

(I'm also thinking we should have a bigger test set that we run before each release to make sure all our examples (and some bigger ones) run and produce the same results as the previous release.)

Here's the command I used that gave an error:

    python -m pdb `which runph` --model-directory=. --instance-directory=inputs/pysp_inputs --solver=gurobi --default-rho=1000.0 --traceback --rho-cfgfile=rhosetter.py  --solution-writer=pyomo.pysp.plugins.csvsolutionwriter --output-scenario-tree-solution

Here were the pdb commands I used to diagnose the error (a little): 

    l 861, 892
    !key, val
    !self._setitem_when_not_present(self._validate_index(key), val)
    !self._validate_index(key)
    !self.name
    !list(self.keys())
    !self.get_index_set().name

josiahjohnston:
# Dependencies

Retiring the inefficient & roundabout sympy parsing in favor of direct Pyomo calls seems great. 

Dropping the direct `numpy` dependency in the `advanced` section won't make a difference because `pandas` is a core dependency and it has a dependency on `numpy`. 

Consolidating the rest of the optional dependencies into a single section called `[extras]` seems fine to me. 

I'm not a fan of making users do ad-hoc installations, and I don't understand benefits of doing that vs registering them explicitly as optional bits in `extras_require`. 

My preference is to keep testfixtures in the core requirements since the testing framework is an integral part of software development, and at this point Switch still seems best-suited for people can do some coding. Hopefully Switch will grow into a broader audience, but in the meanwhile, it seems important to make it easy for people to start developing Switch to expand the base of advanced users & contributors. Moving testfixtures into `[extras]` would be acceptable to me, although I don't understand the problem in having it as a core dependency. 

Maybe you can elaborate on the motivation for reducing dependencies? Design esthetics or some practical problems?

# Installations

The instructions for casual users seems fine to me. For advanced users who will be working with code, I strongly recommend using some kind of virtual environment. Conda's environments seem fine as far as I know, as is virtualenv. I started mucking with pipenv (which seems to be taking over from virtualenv), but I haven't gotten far enough with pipenv to recommend it or write instructions. 

I always recommend using `pip install path/to/src` over `python path/to/src/setup.py install`. Pip has superseded direct execution of setup.py for several years. See [StackOverflow](https://stackoverflow.com/questions/15724093/difference-between-python-setup-py-install-and-pip-install) for a summary of the benefits.

I've played with the idea of setting up a docker image for Switch, but didn't know of a concrete use case. If that ever seems of interest to anyone, I could do that pretty easily.

# Source code access

I like your idea of `switch find <module>`; its implementation should be trivial and it will lower the learning curve for people who aren't yet python gurus. 

I'm not sure how or if the `switch install examples` is better than having users download the examples directory and save it where ever they want. I don't think it would be a big help to casual users who prefer graphical interfaces to command line interfaces, and advanced users who don't mind command line interfaces can do a `cp` directly for greater transparency. Would that command also download the expamples from a corresponding version of the git repo?

Adding something like `--trace [[<module>|<file>[:<function>|<line>]]` argument to `switch solve` seems like a nice idea. I know pdb & ipdb have internal commands for setting arbitrary breakpoints in other files, but I don't know how to interface with them programatically. Ideally, our code would use the most advanced debugger library available in the running python environment (see code snippet below from a different project doing related stuff). Personally, I find the tab-completion & color highlighting in ipdb incredibly useful compared to pdb.

```
def get_debugger_func():
	"""Get a function that will drop the code into an interactive python
	session for debugging and development. Use the most sophisticated
	debugging library that is available (IPython > ipdb > pdb).
	N.B. these will not work in the context of parallel processing.
	"""
	try:
		from IPython import embed
		return embed
	except ImportError:
		from ipdb import set_trace
		return set_trace
	except ImportError:
		from pdb import set_trace
		return set_trace

def main():
	# ...
	debug_func = get_debugger_func()
	debug_func()
```

# rhosetter & Progressive Hedging

The problem was that the input .dat files needed to be regenerated since `VARIABLE_GEN_TPS_RAW` was included in the model. After dealing with that, there were a few bugs in your refactored code from wrong variable names. I fixed them and pushed the updates. It runs to completion now. The expected cost of the objective function doesn't match the value in the README (135,349,802 now vs 111,517,401 in the README), and I don't know if that's due to the solver I used, bug fixes since the last time the README was updated, or some undiagnosed bug. @bmaluenda, would you mind taking a look?

bmaluenda:
I reviewed the PySP example and fixed several bugs, as well as updated the output files to match the format of PySP's current version.

I too am getting objective function values of about 135 MMUSD. I checked the output files and manually estimated the value of the OF and got closer to this value than to the previous ~110. I suspect that we hadn't re-generated the input files and re-run the example since I first uploaded it, so any changes to Switch's OF hadn't been reflected in the output files of this example.

I vote to keep these updated output files and consider them to be correct.

examples/3zone_toy_stochastic_PySP/rhosetter_FS_only.py
```diff
@@ -22,10 +22,10 @@
 optimizations.
 
 """
-from .rhosetter import set_rho_values
+from rhosetter import set_rho_values
```
josiahjohnston: All your other edits look great!
I personally prefer the import style of `.uninstalled_module` for pulling in python files from the same directory that aren't installed anywhere. Whenever I see an `import foo` in python, I tend to assume that foo is some package installed on the system rather than a file in the working directory. Dunno if that's just my personal preference or a wide-spread practice.

josiahjohnston:
Thanks Benjamin!

examples/3zone_toy_stochastic_PySP/rhosetter_FS_only.py
bmaluenda: I removed the relative import and left it as a global import because Python 3 does not support implicit relative imports; you must specify the precise package from where the module will be read (check this [Quora post](https://www.quora.com/Why-does-Python-3-not-allow-relative-imports)). Thus, this solution provides compatibility for both Python 2 and 3.

Maybe you can add a comment specifying that the module is actually another script located in the same folder and not an installed package?

josiahjohnston: Ah. Sounds like I have more learning to do about python 3. Yeah, a comment makes sense in that case.

In general (for other files), this probably means it's extra important to follow to the convention of separating imports into built-in packages, 3rd party packages, and modules/packages from our own codebase.

mfripp: Strangely, there are two places where I can comment on this commit: one in the pull request [(here)](https://github.com/switch-model/switch/pull/115/commits/121fb879abfeca37386cd3c1f4f812a4ca80967c) and one on the commit itself in our repository [(here)](https://github.com/switch-model/switch/commit/121fb879abfeca37386cd3c1f4f812a4ca80967c). Guess this is the better place for it.

Do you happen to know how `rhosetter.py` is found in this case? I tried moving both rhosetter files into a `setters` subdirectory, then using `--rho-cfgfile=setters/rhosetter_FS_only.py`. This worked fine, but I wouldn't actually have expected it to. 

If I break into the code just below this comment, I find that `sys.path` includes `.` (the main directory) but not `''` as defined by [the docs](https://docs.python.org/3/library/sys.html#sys.path) and found from `switch test`, and also not `setters`. Nevertheless, when `from rhosetter import set_rho_values` runs, it manages to find rhosetter.py inside the `setters` directory. I wouldn't expect that to happen.

It looks like `six` (used by Pyomo) is patching the import mechanism, because if I step into the `from rhosetter import set_rho_values` statement, I end up in the `find_module()` function of `six.py`. I didn't follow it all the way through, but maybe `six` is patching Python to support Python-2-style implicit relative imports? That wouldn't be a great situation for us, because this code will break if Pyomo stops using `six`. So maybe the comment should be "rhosetter.py must be in the Python path, e.g. in the directory that is active when `runph` is called."

For what it's worth, I rely fairly often on the fact that the current working directory is at the top of the search path when you call `switch` (or probably `runph`). I find this is a natural way to include custom modules in a particular model. i.e., I create a main directory for a study (a set of scenarios), and I put modules.txt, options.txt, scenarios.txt and any custom modules at the top level. Then I put custom inputs directories for various scenarios inside that (e.g., inputs/fast_rps, inputs/no_wind, etc.). These inputs directories don't have a modules.txt (they share the main one in the top directory), but the scenario definitions in scenarios.txt may use --include-module[s] or --exclude-module[s] to adjust the module list a bit. Then, when I run `switch solve` or `switch solve-scenarios --scenario x`, etc., in the top-level directory, it can automatically find any custom modules that are stored there. Alternatively, if users have a custom modules that they reuse across different studies, they can put them in some other directory and create a .pth file in site-packages that points to that directory (I haven't done that, but I could see it happening).

mfripp: Sorry, following on @josiahjohnston's comment below, I just took another look at `sys.path` from inside `rhosetter_FS_only`, and `setters` is actually there, right at the top (I missed it because it blended in to the line above, and I assumed '.' was the first entry). So what seems to be happening is simply that `runph` adds the location of the `rhosetter_FS_only.py` to the path before importing that module. So as long as `rhosetter.py` is in the same location, it should also load successfully. This is a little brittle (what if `runph` adopts some other technique for importing modules from arbitrary locations?), but it is an attractive behavior, so we could just live with it for now and leave the comment as-is.

josiahjohnston:
Just pushed some more updates that pass all tests in both python 2 & 3:
* `-v`, `-vv` & `-vvv` control level of verbosity via the `logging` python library, and validated its compatibility with old code that used `model.options.verbose` in conjunction with print statements.
    * Extra timepoints for renewable generation or hydro will yield summary warning messages if `-v` is specified, and detailed diagnostic data if `-vv` is specified. 
* Update ArgParser command line definitions to be easier to read on a small screen and easier to search (ex: `--long-arg` gets auto-converted to `options.long_arg`, so including `dest="long_arg"` allows searching between argument definitions & their use in the codebase)
* Minor bug fixes detailed in commit messages.

josiahjohnston:
FYI, I pushed the rebased version of this branch, which I had previously been calling 2.0.6-dev. Leaving it named next_release means we don't have to recreate this pull request, and will be less confusing if we decide to make the next release 2.1.

I updated the outstanding pull requests to be simple differences from the current tip of this branch, and closed an 2-year stale pull request (#97) by dividing into a commit of coding style that is mathematically equivalent, and an issue (#123) for discussin how the formulation is confusing and possibly buggy.

switch_model/solve.py
```diff
@@ -503,8 +504,17 @@ def define_arguments(argparser):
 
     # General purpose arguments
     argparser.add_argument(
-        '--verbose', '-v', dest='verbose', default=False, action='store_true',
+        '--verbose', '-v', dest='verbose', default=False, 
+        action='store_const', const=logging.WARNING,
         help='Show information about model preparation and solution')
+    argparser.add_argument(
```
mfripp: I like the idea of using logging.warn(), logging.info() and logging.debug() and support starting with that here and moving on to eventually using that throughout our code.

However, I think the logging should be more like Python's standard behavior: Logging levels are `critical`, `error`, `warning`, `info`, `debug`. The default is `warning`. For our code, I think the `warning` level should only include things that are most likely errors, that should be reported even if the user isn't looking for extra info. I'm not sure what we would use `critical` or `error` for, since those are probably things that we currently raise errors on. I think the `info` level should provide a running stream of progress messages but not much diagnostics, and the `debug` level should provide more detailed diagnostics.

With that in mind, the `info` level would be equivalent to our current `--verbose` level, and most of our current messages would go there. A few of our messages that currently start with "Warning: ..." would be moved to the warning level. And we can begin adding new debug-level messages as we add new code. 

These levels don't really map onto the flags used in this commit. We could potentially use `--warning`, `--verbose` (info level) and `--very-verbose` (debug level), but the `--warning` flag would be hard to explain (should it really be `--show-warnings`? Should there be a `--no-warning` flag?). Another option would be `--warning`, `--verbose` and `--debug`, but that conflicts with our current debug flag. I would recommend using `--log-level {warning, info, debug}`. That would also fit nicely with the existing `--log-run` and `--logs-dir` flags.

switch_model/generators/extensions/hydro_simple.py
```diff
@@ -88,6 +90,34 @@ def define_components(mod):
...
+            "scheduled retirement date. If you expect those datapoints to "
+            "be useful, then those plants need longer lifetimes (or "
+            "options to build new capacity when the old capacity reaches "
+            "the provided end-of-life date).\n".format(num_impacted_generators))
+        if extra_indexes:
+            logging.warning(warning_msg)
+            logging.info("Plants with extra timepoints:\n{}".format(pprint))
+        return(True)
+    mod.warn_on_extra_HYDRO_GEN_TS = BuildCheck(
+        rule=_warn_on_extra_HYDRO_GEN_TS)
```
mfripp: In https://github.com/switch-model/switch/pull/115/commits/728938566b4552f61366eb4292a3448c6d16658a#r341685397 I recommended using `warning` as our default logging level, and defining the `info` level as being equivalent to our current `--verbose` flag. For the case shown here--where we allow data for extra timepoints but it's not considered good practice for purists--I would show the messages at the `debug` level, not the `info` or `warning` level. If we've decided that this behavior is OK, then we shouldn't hassle the user with warning messages every time they run the model.

switch_model/generators/core/dispatch.py
```diff
@@ -273,6 +274,35 @@ def init_gen_availability(m, g):
...
+            "scheduled retirement date. If you expect those datapoints to "
+            "be useful, then those plants need longer lifetimes (or "
+            "options to build new capacity when the old capacity reaches "
+            "the provided end-of-life date).\n".format(num_impacted_generators))
+        if extra_indexes:
+            logging.warning(warning_msg)
+            logging.info("Plants with extra timepoints:\n{}".format(pprint))
+        return(True)
+    mod.warn_on_extra_VARIABLE_GEN_TPS = BuildCheck(
+        rule=_warn_on_extra_VARIABLE_GEN_TPS)
```
mfripp: In https://github.com/switch-model/switch/pull/115#discussion_r341685397 I recommended using `warning` as our default logging level, and defining the `info` level as being equivalent to our current `--verbose` flag. For the cases shown here--where we allow data for extra timepoints but it's not considered good practice for purists--I would show the messages at the `debug` level, not the `info` or `warning` level. If we've decided that this behavior is OK, then we shouldn't hassle the user with warning messages every time they run the model.

switch_model/generators/core/dispatch.py
```diff
@@ -434,11 +429,31 @@ def post_solve(instance, outdir):
...
-            geom_bar(position="stack") + \
-            scale_y_continuous(name='Energy (GWh/yr)') + theme_bw()
-        annual_summary_plot.save(filename=os.path.join(outdir, "dispatch_annual_summary.pdf"))
+            p9.geom_bar(position="stack") + \
+            p9.scale_y_continuous(name='Energy (GWh/yr)') + \
+            p9.theme_bw()
+        save_as = os.path.join(outdir, "dispatch_annual_summary_tech.pdf")
+        annual_summary_plot.save(filename=save_as)
+    except (ImportError, RuntimeError) as err:
+        pass
```
mfripp: Catching RuntimeError here seems a little broad, e.g., if people want the graph and can't figure out why it's not appearing. How about something more like this:

```
try:
    import plotnine as p9
except ImportError:
    pass
else:
    # all the graphing code
```

switch_model/solve.py
mfripp: Just noticed a lot of this is implemented in commit e5bbfd3b4fb278566f1f81fcc4515802085bc5f4. We may be almost there, but may want to change "--logging-level" to "--log-level" to be more similar to our other flags.

mfripp: I have implemented the approach I recommended above in https://github.com/switch-model/switch/pull/115/commits/29d0332abd8210bd002b0293299ef65762de38a1, but offering `--log-level {error, warning, info, debug}`. We don't offer `--log-level critical`, since that would currently be the same as `--log-level error`. 

Note that my code creates a different logger for each model. This is because each model may have different logging settings, and it is possible for several models to be created during the same Python session (possibly even existing at the same time).

There's more work to do in the future:
- replace all our `print()` calls with `logging.info()`, `logging.warn()`, etc.
- think of a good way to log stderr and stdout to file via the logger instead of using a context handler that tee's these streams. There are a couple of problems with the current approach:
  - output from the solve_scenarios module can't be logged to a file
  - all output logged to file will only go to the most recently specified file

switch_model/generators/core/dispatch.py
mfripp: This is now addressed in https://github.com/switch-model/switch/pull/115/commits/4642de038126b9747cd1ad4f9b6638338b82a7e0. I relented a little and (at the cost of some extra fanciness) give a general message at the `info` level that there are some unused entries in `variable_capacity_factors.csv` (not a warning, just a friendly note), and give more details at the `debug` level. This may be a good template for the future, or may be unnecessarily complex with an unclear line between warning, info and debug.

switch_model/generators/core/dispatch.py
mfripp: This is addressed in https://github.com/switch-model/switch/pull/115/commits/e95abe371f52a73329c78adb8bfe6c48c579f706, along with some other improvements to streamline the code and filter out more warnings.

mfripp:
Closing this because it was actually the 2.0.6 branch, not the next_release branch.

PR #124, opened 2019-08-21
auto_select optional, default to True, unless `select` was explicitly specified

josiahjohnston:
This is an incremental step towards streamlining the load_aug interface. Several modules still relied on explicitly using `select`, and changing those & upgrading inputs would have been a bigger can of worms.


mfripp:
I ended up dropping the `auto_select` argument to `load_aug` entirely, since it no longer has a useful purpose.

Issue #122, opened 2019-08-13
Transmission lines vs. transfer corridors


mfripp:
In switch_model.transmission.transport, we generally refer to entities called "transmission lines". These are really transfer corridors: aggregated inter-zonal transfer capability of the underlying network. Users often estimate this as a derated sum of the ratings of the transmission lines connecting these zones, but there are other ways to do it (e.g., consult the WECC Path Rating Catalog). At any rate, these are definitely not transmission lines, and it could bug electrical engineers to hear them called that. So we should probably rename them as transfer corridors, transmission corridors, transfer paths or similar.

It would also be helpful to add bidirectional ratings for these paths (at least for initial capacity, and possibly different costs to update the capacity in each direction, based on a detailed transmission study). This reflects the fact that transfer paths often have different ratings in each direction.

josiahjohnston:
Ha! This asymmetry issue came up for me a few weeks back. I started looking into implementing it, but my client thought a pathway update might correct the issue. In the handful of datasets I examined before implementing it this way (including WECC), the asymmetries were nominal enough to be attributable to rounding errors. The bi-directionality of Switch 1.0 had also caused many of us significant grief over the years, so dropping that seemed like a nice way forward.

Do you have a clear example of very different capacities and costs for either direction?

I don't have a strong opinion on nomenclature shift to be more precise; the pros & cons in my mind seem fairly balanced, so I'm happy to defer my vote on that. It's up to users whether each zone is a large region or a single bus, so these could represent transmission corridors or single lines. Similarly, a generation project may be a single generator unit or a "virtual plant" that is an aggregation of many different generation plants. 

Anyway, "Transfer Paths" seems fine to me and sufficient ambiguous to handle both use cases. It's probably worth shopping around a little to make sure it's not confusing. @bmaluenda Do you know what language PLEXOS uses for connections in a transport model?

bmaluenda:
I'm not keen on changing the nomenclature. Multiple users (me included) use the Transmission Line objects to model actual lines. The flow model is simply a transport model instead of a DC or AC power flow, but it doesn't change the fact that a line is being represented. I don't think most other users -who use the Trasmission Line objects to model interfaces between zones- would be bothered with the fact of using a "transmission line" obejct to represent a bunch of lines. Josiah's example shows this: many users use the generation unit object to represent a bunch of units (either a plant, a virtual plant, or aggregated capacity in a zone).

[...]

PR #125, opened 2019-08-31
Bidirectional tx builds

josiahjohnston:
This is a working implementation of bi-directional transmission builds, which is a non-controversial and tractable subset of issue #122

As I was working on this, I came up with other items for a wish list, but didn't implement them because they weren't necessary for a minimal viable prototype and I wasn't sure if we would have consensus on their desirability. 
* Consolidate the cost component inputs into a single capital cost per transmission line direction ($/MW-capacity). Have users estimate cost of upgrading corridors using whatever method makes the most sense for them. Provide suggestions based on distance and a default value of $1000/MW-km.
* Rename parameters to use the prefix of tx instead of trans for brevity, and normalize existing_trans_cap to tx_existing_cap
* Investigate shifting TRANSMISSION_LINES to a 2-dimensional set of zone_from, zone_to. If Pyomo gives us the choice of using tx or (z_from,z_to) when iterating & indexing, that would be a win-win for readability and simplicity.


josiahjohnston:
I included another commit to support optional pre-determined transmission builds. It's something I need for a project and seemed uncontroversial, so I'm rolling it into this pull request.

PR #135, opened 2020-12-26
fix typo in planning_reserves

DesmondZhong:
There is a small typo in the file, which is fixed by this commit.


staadecker (approved):

mfripp:
I would call the later commit more of a change in behavior than a fix for errors. As far as I can tell, incremental and total capacity for storage gens were both being written to storage_builds.csv, but this is a request to split the output into storage_builds.csv and storage_capacity.csv to reduce confusion. I'm merging it on that basis, but expect to turn around and immediately combine this with some changes to gen_cap.csv, so there will be a gen_cap.csv and gen_additions.csv, and both of these will have columns for generator power capacity and energy capacity (the latter only applies to storage).

PR #138, opened 2021-04-19
Migrate code style to Black

staadecker:
Hello there!

Through this pull request, I'm proposing that SWITCH use [Black](https://github.com/psf/black), the most used Python code formatter. Here's a description from the Black documentation.

> By using Black, you agree to cede control over minutiae of hand-formatting. In return, Black gives you speed, determinism, and freedom from pycodestyle nagging about formatting. You will save time and mental energy for more important matters.
<br> Black makes code review faster by producing the smallest diffs possible. Blackened code looks the same regardless of the project you’re reading. Formatting becomes transparent after a while and you can focus on the content instead.

Black is managed by the Python Software Foundation, has +20k stars on GitHub and is used by many large projects including the pandas library!

For Switch specifically, it would be great to use Black so that merging changes from downstream forks (e.g. RAEL-Berkley) is easier since formatting would be consistent across both repositories.

This pull request adds black to the dev dependencies and reformats all the Python files according to Black.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,2 @@
+# Migrate code style to Black
+b7131aa93b27e18236880a65f29e36631bb7fb5d
--- a/examples/3zone_toy_stochastic_PySP/PySPInputGenerator.py
+++ b/examples/3zone_toy_stochastic_PySP/PySPInputGenerator.py
@@ -48,6 +48,7 @@ scenario_list should be modified to reflect those changes.
 
 """
 from __future__ import print_function
+
 # Inputs directory relative to the location of this script.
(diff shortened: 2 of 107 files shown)


mfripp:
Just wanted to comment since this has been languishing for a while: I support this idea (though I should probably re-train to use double-quotes instead of single quotes). I expect to start rolling out new versions with elements like this over the next couple months.

staadecker:
@mfripp That sounds great! I'm working with @PatyHidalgo until early August on some research using Switch. We've made quite a few improvements in our fork of Switch and I'd be happy to put in some time to get those changes merged into here. Just let me know once you start working on Switch and we can touch base.

